Issue #6, opened 2022-04-02
Can't set a breakpoint in lambda using expression


jmini:
Given this java source class:

```java
package example;

import java.util.Arrays;
import java.util.List;
import java.util.Optional;

public class Main {

	public static void main(String[] args) {
		Person p1 = new Person("John", "Smith");
		Person p2 = new Person("Bob", "Doe", "1245");
		List<Person> list = Arrays.asList(p1, p2);
		list.stream().forEach(p -> p.getId()
				.ifPresent(id -> {
					System.out.println(p.getFirstName() + " " + p.getLastName()); //<-- line 15
					System.out.println(id);
				}));
		System.out.println("DONE"); //<-- line 18
	}

	private static class Person {
		private String firstName;
		private String lastName;
		private Optional<String> id;

		public Person(String firstName, String lastName) {
			super();
			this.firstName = firstName;
			this.lastName = lastName;
			this.id = Optional.empty();
		}
		public Person(String firstName, String lastName, String id) {
			super();
			this.firstName = firstName;
			this.lastName = lastName;
			this.id = Optional.of(id);
		}

		public String getFirstName() {
			return firstName;
		}

		public String getLastName() {
			return lastName;
		}

		public Optional<String> getId() {
			return id;
		}
	}
}
```

With Eclipse 2022-03, when you try to put a breakpoint in line 15, it is set on line 18.

As demonstrated in this animation:

![Eclipse-breakpoints](https://user-images.githubusercontent.com/1222165/161371730-8319d990-51d7-4f8c-afaf-c02886982260.gif)

---
As a comparison with JDT from Eclipse 2021-12, the breakpoint can be set on the same line.

And the debugger stops at the defined place:

<img width="1059" alt="Screenshot with Eclipse 2021-12" src="https://user-images.githubusercontent.com/1222165/161371786-5d5a882c-e4ba-4c2d-8ad5-7aa02a0785e4.png">

---
With 2022-03, if the code is modified to convert the first lambda to a block instead of an expression:

```java
	public static void main(String[] args) {
		Person p1 = new Person("John", "Smith");
		Person p2 = new Person("Bob", "Doe", "1245");
		List<Person> list = Arrays.asList(p1, p2);
		list.stream().forEach(p -> {
			p.getId()
					.ifPresent(id -> {
						System.out.println(p.getFirstName() + " " + p.getLastName()); //<-- line 16
						System.out.println(id);
					});
		});
[...]

PR #7, opened 2022-04-04
Can't set a breakpoint in lambda using expression #6

iloveeclipse:
Check nested method invocations for nested lambda bodies.
Fixes issue #6.


iloveeclipse:
@SarikaSinha : I'm not sure if the fix is right or not, I'm not experienced in AST7DOm traversal. At least given example is working now.

PR #14, opened 2022-04-07
Bug 547041 - Fixed parsing of Debug view stacks with multiple generics

trancexpress:
If the type in a Debug view stack trace line has multiple type
parameters, the regexp for hyperlinks does not correctly detect the
whole line.

This change adds also comma to the accepted symbols for the type name
part of stack trace lines.

Change-Id: I31d38f2679caafa63c765efeb39d80ea831624cb
Signed-off-by: Simeon Andreev <simeon.danailov.andreev@gmail.com>


iloveeclipse (approved):
LGTM. 
Let's wait for M1 to be declared before merging.

iloveeclipse:
Verified with I20220410-1800

PR #42, opened 2022-05-09
Create codeql-analysis and status on github project page

carstenartur:


SarikaSinha:
@carstenartur 
Thank you for taking the initiative.
I will suggest to start a discussion on the common approach in general for all the eclipse platform projects before taking this forward.  Also it will be good to have separate issues to handle a new type of dependency/check like code coverage, vulnerability etc and release one by one after one change becomes stable.

PR #43, opened 2022-05-10
Bug 575131 - Deadlock with conditional breakpoints that suspend VM

trancexpress:
Bug 575131 - Deadlock with conditional breakpoints that suspend VM

Adjusts JDIDebugTarget resumeThreads() and
suspendThreads() to not resume resp. suspend threads that are handling
breakpoints.

Also adds a test to reproduce the deadlock, as well as a stress test
suite that will repeat the new test (since the deadlock is sporadic).
The stress test suite is not executed during automated tests.

Change-Id: Icfcd5ebb04e77e437bc308d5cb6911a77eb3b673
Signed-off-by: Simeon Andreev <simeon.danailov.andreev@gmail.com>

### DIFF
--- a/org.eclipse.jdt.debug.tests/testprograms/SuspendVMConditionalBreakpointsTestSnippet.java
+++ b/org.eclipse.jdt.debug.tests/testprograms/SuspendVMConditionalBreakpointsTestSnippet.java
@@ -1,0 +1,45 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Simeon Andreev and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Simeon Andreev - initial API and implementation
+ *******************************************************************************/
+
+/**
+ * Snippet for: Bug 575131 - Deadlock during "JDI Expression Evaluation Event Dispatch"
+ */
+public class SuspendVMConditionalBreakpointsTestSnippet {
+
+	public static final int i = 50;
+	public static final int j = 25;
+
+	public static void main(String[] args) throws InterruptedException {
+		Thread t1 = new Thread(new Runnable() {
+			@Override
+			public void run() {
+				for (int k = 0; k < 5; ++k) {
+					System.out.println("thread i=" + i); // suspend VM breakpoint with condition "if (i == 50) { System.out.println(i); } return false;"
+				}
+			}
+		});
+		Thread t2 = new Thread(new Runnable() {
+			@Override
+			public void run() {
+				for (int k = 0; k < 5; ++k) {
+					System.out.println("thread j=" + j); // suspend VM breakpoint with condition "if (j == 25) { System.out.println(j); } return false;"
+				}
+			}
+		});
+		t1.start();
+		t2.start();
+		t1.join();
+		t2.join();
+	}
+}
--- a/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/StressTestSuspendVMConditionalBreakpoints.java
+++ b/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/StressTestSuspendVMConditionalBreakpoints.java
@@ -1,0 +1,37 @@
+/*******************************************************************************
+ *  Copyright (c) 2022 Simeon Andreev and others.
+ *
+ *  This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License 2.0
+ *  which accompanies this distribution, and is available at
+ *  https://www.eclipse.org/legal/epl-2.0/
+ *
+ *  SPDX-License-Identifier: EPL-2.0
+ *
+ *  Contributors:
+ *     Simeon Andreev - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.jdt.debug.tests;
+
+import org.eclipse.jdt.debug.tests.breakpoints.SuspendVMConditionalBreakpointsTests;
+
+import junit.framework.Test;
+import junit.framework.TestSuite;
+
+/**
+ * Stress test suite for bug 575131. Not executed in automated tests.
+ */
+public class StressTestSuspendVMConditionalBreakpoints extends DebugSuite {
+
+	private static final int STRESS_TEST_REPEAT_COUNT = 500;
+
+	public static Test suite() {
+		return new StressTestSuspendVMConditionalBreakpoints();
+	}
+
+	public StressTestSuspendVMConditionalBreakpoints() {
+		for (int i = 0; i < STRESS_TEST_REPEAT_COUNT; ++i) {
+			addTest(new TestSuite(SuspendVMConditionalBreakpointsTests.class));
+		}
+	}
+}
--- a/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/breakpoints/SuspendVMConditionalBreakpointsTests.java
+++ b/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/breakpoints/SuspendVMConditionalBreakpointsTests.java
@@ -1,0 +1,82 @@
+/*******************************************************************************
+ *  Copyright (c) 2022 Simeon Andreev and others.
+ *
+ *  This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License 2.0
+ *  which accompanies this distribution, and is available at
+ *  https://www.eclipse.org/legal/epl-2.0/
+ *
+ *  SPDX-License-Identifier: EPL-2.0
+ *
+ *  Contributors:
+ *     Simeon Andreev - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.jdt.debug.tests.breakpoints;
+
+import org.eclipse.debug.core.DebugEvent;
+import org.eclipse.debug.core.ILaunchConfiguration;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.debug.core.IJavaBreakpoint;
+import org.eclipse.jdt.debug.core.IJavaDebugTarget;
+import org.eclipse.jdt.debug.core.IJavaLineBreakpoint;
+import org.eclipse.jdt.debug.core.IJavaThread;
+import org.eclipse.jdt.debug.testplugin.DebugElementKindEventWaiter;
+import org.eclipse.jdt.debug.testplugin.DebugEventWaiter;
+import org.eclipse.jdt.debug.tests.AbstractDebugTest;
+import org.eclipse.jdt.debug.tests.TestUtil;
+
+/**
+ * Test that a SUSPEND_VM conditional breakpoints function as expected. Also tests for bug 575131.
+ */
+public class SuspendVMConditionalBreakpointsTests extends AbstractDebugTest {
+
+	/**
+	 * Constructor
+	 *
+	 * @param name
+	 */
+	public SuspendVMConditionalBreakpointsTests(String name) {
+		super(name);
+	}
+
+	/**
+	 * Tests that the VM is suspended when a conditional line breakpoint is hit.
+	 */
+	public void testSuspendVmLineConditionalBreakpoint() throws Exception {
+		String typeName = "SuspendVMConditionalBreakpointsTestSnippet";
+
+		IJavaLineBreakpoint bp1 = createLineBreakpoint(28, typeName);
+		bp1.setCondition("if (i == 50) { System.out.println(i); } return false;");
+		bp1.setConditionEnabled(true);
+		bp1.setSuspendPolicy(IJavaBreakpoint.SUSPEND_VM);
+		IJavaLineBreakpoint bp2 = createLineBreakpoint(36, typeName);
+		bp2.setCondition("if (j == 25) { System.out.println(j); } return false;");
+		bp2.setConditionEnabled(true);
+		bp2.setSuspendPolicy(IJavaBreakpoint.SUSPEND_VM);
+
+		IJavaDebugTarget debugTarget = null;
+		try {
+			TestUtil.waitForJobs("testSuspendVmLineConditionalBreakpoint before launch", 250, 500);
+			IJavaProject project = getProjectContext();
+			ILaunchConfiguration config = getLaunchConfiguration(project, typeName);
+
+			DebugEventWaiter waiter = new DebugElementKindEventWaiter(DebugEvent.BREAKPOINT, IJavaThread.class);
+			waiter.setTimeout(DEFAULT_TIMEOUT);
+			waiter.setEnableUIEventLoopProcessing(enableUIEventLoopProcessingInWaiter());
+
+			boolean registerLaunch = true;
+			IJavaThread suspendedThread = (IJavaThread) launchAndWait(config, waiter, registerLaunch);
+
+			debugTarget = (IJavaDebugTarget) suspendedThread.getDebugTarget();
+			TestUtil.waitForJobs("testSuspendVmLineConditionalBreakpoint after suspend", 250, 500);
+
+			debugTarget.resume();
+			TestUtil.waitForJobs("testSuspendVmLineConditionalBreakpoint after resume", 250, 500);
+		} finally {
+			if (debugTarget != null) {
+				terminateAndRemove(debugTarget);
+			}
+			removeAllBreakpoints();
+		}
+	}
+}
--- a/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java
+++ b/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java
@@ -1733,7 +1733,10 @@ public class JDIDebugTarget extends JDIDebugElement implements
 	protected void suspendThreads() {
 		Iterator<JDIThread> threads = getThreadIterator();
 		while (threads.hasNext()) {
-			threads.next().suspendedByVM();
+			JDIThread thread = threads.next();
+			if (!thread.isBreakpointHandlingOngoing()) {
+				thread.suspendedByVM();
+			}
 		}
 	}
 
@@ -1743,7 +1746,10 @@ public class JDIDebugTarget extends JDIDebugElement implements
 	protected void resumeThreads() throws DebugException {
 		Iterator<JDIThread> threads = getThreadIterator();
 		while (threads.hasNext()) {
-			threads.next().resumedByVM();
+			JDIThread thread = threads.next();
+			if (!thread.isBreakpointHandlingOngoing()) {
+				thread.resumedByVM();
+			}
 		}
 	}
 
--- a/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIThread.java
+++ b/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIThread.java
@@ -319,6 +319,9 @@ public class JDIThread extends JDIDebugElement implements IJavaThread {
 	 */
 	private volatile IJavaObject fPreviousException;
 
+	private final AtomicBoolean fCompletingBreakpointHandling;
+	private final AtomicBoolean fHandlingSuspendForBreakpoint;
+
 	/**
 	 * Creates a new thread on the underlying thread reference in the given
 	 * debug target.
@@ -337,6 +340,8 @@ public class JDIThread extends JDIDebugElement implements IJavaThread {
 		setUnderlyingThread(thread);
 		fAsyncJob = new ThreadJob();
 		initialize();
+		fCompletingBreakpointHandling = new AtomicBoolean(false);
+		fHandlingSuspendForBreakpoint = new AtomicBoolean(false);
 	}
 
 	/**
@@ -1358,6 +1363,15 @@ public class JDIThread extends JDIDebugElement implements IJavaThread {
 	 */
 	public boolean handleSuspendForBreakpoint(JavaBreakpoint breakpoint,
 			boolean suspendVote) {
+		fHandlingSuspendForBreakpoint.set(true);
+		try {
+			return handleSuspendForBreakpointInternal(breakpoint);
+		} finally {
+			fHandlingSuspendForBreakpoint.set(true);
+		}
+	}
+
+	private boolean handleSuspendForBreakpointInternal(JavaBreakpoint breakpoint) {
 		int policy = IJavaBreakpoint.SUSPEND_THREAD;
 		synchronized (this) {
 			if (fClientSuspendRequest) {
@@ -1381,6 +1395,7 @@ public class JDIThread extends JDIDebugElement implements IJavaThread {
 			if (policy == IJavaBreakpoint.SUSPEND_VM) {
 				((JDIDebugTarget) getDebugTarget())
 						.prepareToSuspendByBreakpoint(breakpoint);
+				suspendedByVM();
 			} else {
 				setRunning(false);
 			}
@@ -1495,6 +1510,15 @@ public class JDIThread extends JDIDebugElement implements IJavaThread {
 	 */
 	public void completeBreakpointHandling(JavaBreakpoint breakpoint,
 			boolean suspend, boolean queue, EventSet set) {
+		fCompletingBreakpointHandling.set(true);
+		try {
+			completeBreakpointHandlingInternal(breakpoint, suspend, queue, set);
+		} finally {
+			fCompletingBreakpointHandling.set(false);
+		}
+	}
+
+	private void completeBreakpointHandlingInternal(JavaBreakpoint breakpoint, boolean suspend, boolean queue, EventSet set) {
 		synchronized (this) {
 			try {
 				int policy = breakpoint.getSuspendPolicy();
@@ -1514,6 +1538,7 @@ public class JDIThread extends JDIDebugElement implements IJavaThread {
 					if (policy == IJavaBreakpoint.SUSPEND_VM) {
 						((JDIDebugTarget) getDebugTarget())
 								.cancelSuspendByBreakpoint(breakpoint);
+						resumedByVM();
 					} else {
 						setRunning(true);
 						// dispose cached stack frames so we re-retrieve on the
@@ -1526,7 +1551,6 @@ public class JDIThread extends JDIDebugElement implements IJavaThread {
 				setRunning(true);
 			}
 		}
-
 	}
 
 	@Override
@@ -3649,4 +3673,7 @@ public class JDIThread extends JDIDebugElement implements IJavaThread {
 		this.fMethodResult = fMethodResult;
 	}
 
+	boolean isBreakpointHandlingOngoing() {
+		return fCompletingBreakpointHandling.get() || fHandlingSuspendForBreakpoint.get();
+	}
 }
--- a/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIThread.java
+++ b/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIThread.java
@@ -1367,7 +1367,7 @@ public class JDIThread extends JDIDebugElement implements IJavaThread {
 		try {
 			return handleSuspendForBreakpointInternal(breakpoint);
 		} finally {
-			fHandlingSuspendForBreakpoint.set(true);
+			fHandlingSuspendForBreakpoint.set(false);
 		}
 	}
 
--- a/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/AbstractDebugTest.java
+++ b/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/AbstractDebugTest.java
@@ -41,6 +41,7 @@ import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.Platform;
@@ -197,7 +198,8 @@ public abstract class AbstractDebugTest extends TestCase implements  IEvaluation
 			"org.eclipse.debug.tests.targets.HcrClass9", "TestContributedStepFilterClass", "TerminateAll_01", "TerminateAll_02", "StepResult1",
 			"StepResult2", "StepResult3", "StepUncaught", "TriggerPoint_01", "BulkThreadCreationTest", "MethodExitAndException",
 			"Bug534319earlyStart", "Bug534319lateStart", "Bug534319singleThread", "Bug534319startBetwen", "MethodCall", "Bug538303", "Bug540243",
-			"OutSync", "OutSync2", "ConsoleOutputUmlaut", "ErrorRecurrence", "ModelPresentationTests", "Bug565982" };
+			"OutSync", "OutSync2", "ConsoleOutputUmlaut", "ErrorRecurrence", "ModelPresentationTests", "Bug565982",
+			"SuspendVMConditionalBreakpointsTestSnippet" };
 
 	/**
 	 * the default timeout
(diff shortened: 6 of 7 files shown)


org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/StressTestSuspendVMConditionalBreakpoints.java
```diff
@@ -0,0 +1,21 @@
+package org.eclipse.jdt.debug.tests;
```
trancexpress, line 1: Will likely be kicking out this file from the final version of the commit, so no copyright header.

Also 500 repeats is exaggerated, I got a deadlock already in the first few test runs.

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/breakpoints/SuspendVMConditionalBreakpointsTests.java
```diff
@@ -0,0 +1,103 @@
...
+		bp1.setConditionEnabled(true);
+		bp1.setSuspendPolicy(IJavaBreakpoint.SUSPEND_VM);
+		IJavaLineBreakpoint bp2 = createLineBreakpoint(38, typeName);
+		bp2.setCondition("if (j == 25) { System.out.println(j); } return false;");
+		bp2.setConditionEnabled(true);
+		bp2.setSuspendPolicy(IJavaBreakpoint.SUSPEND_VM);
+
+		IJavaDebugTarget debugTarget = null;
+		try {
+			TestUtil.waitForJobs("testSuspendVmLineConditionalBreakpoint before launch", 250, 500);
```
trancexpress, line 62: The waits are probably not needed. Will have to check whether the deadlock is reproducible without them.

trancexpress:
So far no clue how to fix this.

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/breakpoints/SuspendVMConditionalBreakpointsTests.java
trancexpress: > The waits are probably not needed. Will have to check whether the deadlock is reproducible without them.

Seems like the waits help with faster reproduction. The only wait I could remove and still run into the deadlock in a few test executions, was this one:

```
diff --git a/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/breakpoints/SuspendVMConditionalBreakpointsTests.java b/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/breakpoints/SuspendVMConditionalBreakpointsTests.java
index 32f9ec7e7..4dd96f8cc 100644
--- a/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/breakpoints/SuspendVMConditionalBreakpointsTests.java
+++ b/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/breakpoints/SuspendVMConditionalBreakpointsTests.java
@@ -69,7 +69,6 @@ public class SuspendVMConditionalBreakpointsTests extends AbstractDebugTest {
 
                        boolean registerLaunch = true;
                        launchAndWait(config, waiter, registerLaunch);
-                       TestUtil.waitForJobs("testSuspendVmLineConditionalBreakpoint after launch", 250, 500);
 
                        debugTarget = getDebugTarget();
                        TestUtil.waitForJobs("testSuspendVmLineConditionalBreakpoint after suspend", 250, 500);
```

trancexpress:
No hangs overnight with looping the test added here.

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/breakpoints/SuspendVMConditionalBreakpointsTests.java
```diff
@@ -0,0 +1,101 @@
...
+	 * Constructor
+	 *
+	 * @param name
+	 */
+	public SuspendVMConditionalBreakpointsTests(String name) {
+		super(name);
+	}
+
+	/**
+	 * Tests that the VM is suspended when the a conditional line breakpoint is hit.
```
iloveeclipse, line 45: Not sure I understood what the test actually does & when it should fail according to the comment one thread & one breakpoint are only needed. But I see **two** threads & **two** breakpoints? 
I guess the comment should be updated and explain that we test **race conditions** while setting VM suspending breakpoints.

trancexpress: https://bugs.eclipse.org/bugs/show_bug.cgi?id=575131#c3

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/breakpoints/SuspendVMConditionalBreakpointsTests.java
```diff
@@ -0,0 +1,101 @@
...
+			ILaunchConfiguration config = getLaunchConfiguration(project, typeName);
+
+			DebugEventWaiter waiter = new DebugElementKindEventWaiter(DebugEvent.BREAKPOINT, IJavaThread.class);
+			waiter.setTimeout(DEFAULT_TIMEOUT);
+			waiter.setEnableUIEventLoopProcessing(enableUIEventLoopProcessingInWaiter());
+
+			boolean registerLaunch = true;
+			launchAndWait(config, waiter, registerLaunch);
+
+			debugTarget = getDebugTarget();
```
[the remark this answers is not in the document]
iloveeclipse: > At most we can ask the returned thread for its debug target.

Isn't that exactly what I've explained above?

org.eclipse.jdt.debug.tests/testprograms/SuspendVMConditionalBreakpointsTestSnippet.java
```diff
@@ -0,0 +1,51 @@
...
+	public static final int i = 50;
+	public static final int j = 25;
+
+	public static void main(String[] args) throws InterruptedException {
+
+		Thread t1 = new Thread(new Runnable() {
+			@Override
+			public void run() {
+				for (int k = 0; k < 5; ++k) {
+					System.out.println("thread i=" + i + ", k=" + k);
```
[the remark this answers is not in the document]
trancexpress: I think I added setting the thread name after creating the test, I'll adjust the line.

org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java
```diff
@@ -1781,8 +1784,9 @@ protected void suspendedByBreakpoint(JavaBreakpoint breakpoint,
 	 */
 	protected void cancelSuspendByBreakpoint(JavaBreakpoint breakpoint)
 			throws DebugException {
-		setSuspended(false);
-		resumeThreads();
+		if (suspended.getAndSet(false)) {
```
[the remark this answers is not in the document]
trancexpress: Yes, ```suspended``` is left-over from previous code. It should be ```fSuspended```.

There is ```SuspendVMBreakpointsTests``` but I guess no test there checks whether this method works correctly (in particular, checking whether calling this method twice works on the same debug target).

I can try to add a test for it. In this issue, or in a new one? No idea how long it will take to write the test.

trancexpress:
> What I miss is the deadlock **analysis**, either on the issue or in the commit message.

Which issue? There is only https://bugs.eclipse.org/bugs/show_bug.cgi?id=575131.

I'll see if I can fit the description of the deadlock in the commit message.

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/breakpoints/SuspendVMConditionalBreakpointsTests.java
iloveeclipse: You're asked for review, I did it, and in order to understand the change I need explanation what the test is supposed to proof and how that is related to the bug.

iloveeclipse:
In https://bugs.eclipse.org/bugs/show_bug.cgi?id=575131#c1  I wrote:

> I wonder why do we have *two* "JDI Expression Evaluation Event Dispatch" threads?
> org.eclipse.jdt.internal.debug.core.EventDispatcher javadoc says:
> "There is one event dispatcher per JDI debug target."

Have you looked into this? Assuming the events would be not handled by multiple dispatch threads, we would not have the race condition here.

trancexpress:
> Have you looked into this? Assuming the events would be not handled by multiple dispatch threads, we would not have the race condition here.

Sorry its been a while, I forgot about that question. I'll look into it.

trancexpress:
OK after fixing the ```suspended``` vs. ```fSuspended``` setting, another deadlock occurs (as expected, suspending all threads is also a problem):

```
Java stack information for the threads listed above:
===================================================
"Worker-4: Label Job":
        at org.eclipse.jdt.internal.debug.core.model.JDIThread.isOutOfSynch(JDIThread.java:1584)
        - waiting to lock <0x00000000f90058a8> (a org.eclipse.jdt.internal.debug.core.model.JDIThread)
        at org.eclipse.jdt.internal.debug.core.model.JDIDebugTarget.isOutOfSynch(JDIDebugTarget.java:986)
        at org.eclipse.jdt.internal.debug.ui.JDIModelPresentation.computeJDIAdornmentFlags(JDIModelPresentation.java:1052)
        at org.eclipse.jdt.internal.debug.ui.JDIModelPresentation.getDebugElementImage(JDIModelPresentation.java:995)
        at org.eclipse.jdt.internal.debug.ui.JDIModelPresentation.getImage(JDIModelPresentation.java:731)
        at org.eclipse.debug.internal.ui.LazyModelPresentation.getImage(LazyModelPresentation.java:129)
        at org.eclipse.debug.internal.ui.DelegatingModelPresentation.getImage(DelegatingModelPresentation.java:138)
        at org.eclipse.debug.internal.ui.views.launch.DebugElementHelper.getImageDescriptor(DebugElementHelper.java:73)
        at org.eclipse.debug.internal.ui.model.elements.DebugElementLabelProvider.getImageDescriptor(DebugElementLabelProvider.java:80)
        at org.eclipse.debug.internal.ui.model.elements.ElementLabelProvider.getImageDescriptor(ElementLabelProvider.java:274)
        at org.eclipse.debug.internal.ui.model.elements.ElementLabelProvider.retrieveLabel(ElementLabelProvider.java:201)
        at org.eclipse.debug.internal.ui.model.elements.ElementLabelProvider$LabelUpdater.run(ElementLabelProvider.java:147)
        at org.eclipse.debug.internal.ui.model.elements.ElementLabelProvider$LabelJob.run(ElementLabelProvider.java:74)
        at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
"Worker-12: JDI Expression Evaluation Event Dispatch":
        at org.eclipse.jdt.internal.debug.core.model.JDIThread.resumedByVM(JDIThread.java:2008)
        - waiting to lock <0x00000000f9007588> (a org.eclipse.jdt.internal.debug.core.model.JDIThread)
        at org.eclipse.jdt.internal.debug.core.model.JDIDebugTarget.resumeThreads(JDIDebugTarget.java:1747)
        at org.eclipse.jdt.internal.debug.core.model.JDIDebugTarget.cancelSuspendByBreakpoint(JDIDebugTarget.java:1786)
        at org.eclipse.jdt.internal.debug.core.model.JDIThread.completeBreakpointHandling(JDIThread.java:1516)
        - locked <0x00000000f90058a8> (a org.eclipse.jdt.internal.debug.core.model.JDIThread)
        at org.eclipse.jdt.internal.debug.core.breakpoints.JavaBreakpoint.eventSetComplete(JavaBreakpoint.java:345)
        at org.eclipse.jdt.internal.debug.core.EventDispatcher.dispatch(EventDispatcher.java:211)
        at org.eclipse.jdt.internal.debug.core.EventDispatcher$1.run(EventDispatcher.java:311)
        at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
"Worker-8: JDI Expression Evaluation Event Dispatch":
        at org.eclipse.jdt.internal.debug.core.model.JDIThread.suspendedByVM(JDIThread.java:1998)
        - waiting to lock <0x00000000f90058a8> (a org.eclipse.jdt.internal.debug.core.model.JDIThread)
        at org.eclipse.jdt.internal.debug.core.model.JDIDebugTarget.suspendThreads(JDIDebugTarget.java:1737)
        at org.eclipse.jdt.internal.debug.core.model.JDIDebugTarget.prepareToSuspendByBreakpoint(JDIDebugTarget.java:1759)
        at org.eclipse.jdt.internal.debug.core.model.JDIThread.handleSuspendForBreakpoint(JDIThread.java:1383)
        - locked <0x00000000f9007588> (a org.eclipse.jdt.internal.debug.core.model.JDIThread)
        at org.eclipse.jdt.internal.debug.core.breakpoints.JavaBreakpoint.suspend(JavaBreakpoint.java:410)
        at org.eclipse.jdt.internal.debug.core.breakpoints.JavaBreakpoint.handleBreakpointEvent(JavaBreakpoint.java:399)
        at org.eclipse.jdt.internal.debug.core.breakpoints.JavaBreakpoint.handleEvent(JavaBreakpoint.java:315)
        at org.eclipse.jdt.internal.debug.core.EventDispatcher.dispatch(EventDispatcher.java:190)
        at org.eclipse.jdt.internal.debug.core.EventDispatcher$1.run(EventDispatcher.java:311)
        at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

Anyway I'm looking into the multiple dispatchers.

trancexpress:
There is a job that is scheduled from the dispatcher, it was added for: https://bugs.eclipse.org/bugs/show_bug.cgi?id=285130

```
// 269231 always evaluate expressions in a separate job to avoid deadlocks
Job job = new AbstractDispatchJob("JDI Expression Evaluation Event Dispatch") { //$NON-NLS-1$
	@Override
	protected IStatus run(IProgressMonitor monitor) {
		dispatch(eventSet);
		return Status.OK_STATUS;
	}
};
job.setSystem(true);
job.schedule();
```

I assume this is why we see the deadlock with conditional breakpoints only.

Do we want a mutually exclusive rule for this job? Should hopefully be fine to let evaluations evaluate 1 at a time. Unless some evaluation takes a while to finish?...

iloveeclipse:
OK, I read comments on original patch here: https://bugs.eclipse.org/bugs/show_bug.cgi?id=285130#c39 and here: https://bugs.eclipse.org/bugs/show_bug.cgi?id=285130#c41 and see my concerns regarding event serialization pointing to  https://bugs.eclipse.org/bugs/show_bug.cgi?id=269231

Tried the patch below and hangs with modified example from https://bugs.eclipse.org/bugs/show_bug.cgi?id=269231.
```
public class BreakpointListenerTest {

		foo(); // conditional breakpoint here: foo(); return false;
		System.out.println("out of foo");
		bar();
		System.out.println("out of bar");
	}

	private static void foo() {
		System.out.println("hello foo"); // breakpoint here
		bar(); // hangs debugger (evaluation triggers evaluation)
	}

	private static void bar() {
		System.out.println("hello bar"); // conditional breakpoint here: bar(); return false;
	}
}
```
Path that breaks nested evaluations:

```
diff --git a/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/EventDispatcher.java b/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/EventDispatcher.java
index 48e1fe1..266c1de 100644
--- a/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/EventDispatcher.java
+++ b/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/EventDispatcher.java
@@ -25,2 +25,3 @@
 import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.ISchedulingRule;
 import org.eclipse.core.runtime.jobs.Job;
@@ -314,2 +315,3 @@
 							};
+							job.setRule(new ExclusiveJobRule());
 							job.setSystem(true);
@@ -325,2 +327,15 @@
 
+	static class ExclusiveJobRule implements ISchedulingRule {
+
+		@Override
+		public boolean contains(ISchedulingRule rule) {
+			return rule instanceof ExclusiveJobRule;
+		}
+
+		@Override
+		public boolean isConflicting(ISchedulingRule rule) {
+			return rule instanceof ExclusiveJobRule;
+		}
+
+	}
 	/**
```

trancexpress:
> Tried the patch below and it hangs with modified example from https://bugs.eclipse.org/bugs/show_bug.cgi?id=269231.

Should I simplify the reproduction snippet in the test then?

> Patch that breaks nested evaluations:

I was trying this out... unfortunate that nested evaluations don't work. Do we continue with the current proposal then? It needs the same treatment for the suspend all threads path.

iloveeclipse:
Yes and yes.

iloveeclipse:
I wonder if we should serialize calls that do VM suspend/resume by a dedicated job/rule.
- JDIThread.completeBreakpointHandling(JDIThread.java:1516)
- JDIThread.handleSuspendForBreakpoint(JDIThread.java:1383)

trancexpress:
> I wonder if we should serialize calls that do VM suspend/resume by a dedicated job/rule.
> 
>     * JDIThread.completeBreakpointHandling(JDIThread.java:1516)
> 
>     * JDIThread.handleSuspendForBreakpoint(JDIThread.java:1383)
> 
> 
> means, do the concrete BP event processing as today except that trigger **async** job with the exclusive rule to suspend/resume JVM.

If we can find an alternative route, that would be great. The debugger code already has way too many layers of async execution.

It might of course be our only option, if it doesn't break anything... I'm not sure how we want to sync thread resuming and suspending, on breakpoint handling start resp. finish... https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/43#issuecomment-1124610216

org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java
```diff
@@ -1754,8 +1755,9 @@ protected void resumeThreads() throws DebugException {
 	 *            the breakpoint that caused the suspension
 	 */
 	public void prepareToSuspendByBreakpoint(JavaBreakpoint breakpoint) {
-		setSuspended(true);
-		suspendThreads();
+		if (!fSuspended.getAndSet(true)) {
```
trancexpress, line 1758: I assume this will not be enough, but I'm running the reproducer test just to be sure.

trancexpress:
No deadlock overnight, 15k test repetitions. I'll look into the suspend/resume stack trace to maybe understand why it doesn't occur now.

trancexpress:
No hangs with the asnyc job, as expected. Still, probably not the best approach (if there is an alternative).

org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java
```diff
@@ -1755,7 +1756,29 @@ protected void resumeThreads() throws DebugException {
 	 */
 	public void prepareToSuspendByBreakpoint(JavaBreakpoint breakpoint) {
 		setSuspended(true);
-		suspendThreads();
+		Job job = new Job("Suspend all threads from prepareToSuspendByBreakpoint") { //$NON-NLS-1$
```
iloveeclipse, line 1759: OK, now the first question is: what if we have **N > 2** suspendAll / resumeAll jobs scheduled?
Does it make sense to run them **all** serialized? I guess no.

Note: because the execution order is **not guaranteed** by current code, and also it is impossible to guarantee it also if we would like to do so, we are free to discard all but the last job from the queue (so one is running, one is waiting to run).

Something like `org.eclipse.debug.internal.ui.model.elements.SerialExecutor` with a size bounded (max two) queue?

Ideally also this job would be managed by `EventDispatcher` so debug target can forward the operation like `getEventDispatcher().suspendAllThreads()`. This is to allow the job to be terminated / managed by `EventDispatcher`.

trancexpress:
I'm not sure the right place is ```EventDispatcher```, it would result in calls that go ```JDIDebugTarget``` -> ```EventDispatcher``` -> ```JDIDebugTarget```.

We can anyway discuss later on, I'll focus on allowing only the last suspend resp. resume to be called.

iloveeclipse:
> I'm not sure the right place is `EventDispatcher`, it would result in calls that go `JDIDebugTarget` -> `EventDispatcher` -> `JDIDebugTarget`.

That happens already on many events:
image

iloveeclipse:
There is one related test fail:
```
junit.framework.AssertionFailedError: Thread wasn't suspended when a SUSPEND_VM breakpoint was hit, thread=main
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.assertTrue(Assert.java:22)
	at junit.framework.TestCase.assertTrue(TestCase.java:192)
	at org.eclipse.jdt.debug.tests.breakpoints.SuspendVMBreakpointsTests.verifyAllThreadsSuspended(SuspendVMBreakpointsTests.java:215)
	at org.eclipse.jdt.debug.tests.breakpoints.SuspendVMBreakpointsTests.testSuspendVmMethodExitBreakpoint(SuspendVMBreakpointsTests.java:197)
```

trancexpress:
> junit.framework.AssertionFailedError: Thread wasn't suspended when a SUSPEND_VM breakpoint was hit, thread=main
> 	at junit.framework.Assert.fail(Assert.java:57)
> 	at junit.framework.Assert.assertTrue(Assert.java:22)
> 	at junit.framework.TestCase.assertTrue(TestCase.java:192)
> 	at org.eclipse.jdt.debug.tests.breakpoints.SuspendVMBreakpointsTests.verifyAllThreadsSuspended(SuspendVMBreakpointsTests.java:215)
> 	at org.eclipse.jdt.debug.tests.breakpoints.SuspendVMBreakpointsTests.testSuspendVmMethodExitBreakpoint(SuspendVMBreakpointsTests.java:197)

Should be due to the extra added indirection. For the new test this is irrelevant, I added some check so that the test actually tests something. But we only care about having no hangs when running the test. That some thread was suspended slightly later on, is probably not relevant for the user.

We might of course introduce other sporadic fails.

trancexpress:
> I guess what we could do (the test here would still fail sporadically) is introduce new methods to `JDIDebugTarget`, "suspend/resume threads in a job". That way we don't change existing functionality, other than the parts that are deadlocked here.

Thinking about this, it might not be a good idea either. Both relevant methods in ```JDIDebugTarget``` can be overriden in a subtype and the subtype might expect those calls on thread suspend/resume...

trancexpress:
> For the new test this is irrelevant, I added some check so that the test actually tests something. But we only care about having no hangs when running the test. That some thread was suspended slightly later on, is probably not relevant for the user.

I got confused with the new test. This is actually an existing test. So we do introduce a regression with the latest change, at least for code that expects suspended threads after breakpoint handling. Like in the failed test.

iloveeclipse:
I think test should be adopted. We can't fix deadlock and hold the promise that all threads will be immediately suspended /resumed after breakpoint handling code.

trancexpress:
I still saw a sporadic fail in ```SuspendVMBreakpointsTests.testSuspendVmMethodExitBreakpoint()```, despite the new wait. The failed condition was:

```
junit.framework.AssertionFailedError: suspended, but not by breakpoint
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.assertTrue(Assert.java:22)
	at junit.framework.Assert.assertNotNull(Assert.java:256)
	at junit.framework.TestCase.assertNotNull(TestCase.java:399)
	at org.eclipse.jdt.debug.tests.breakpoints.SuspendVMBreakpointsTests.testSuspendVmMethodExitBreakpoint(SuspendVMBreakpointsTests.java:194)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at junit.framework.TestCase.runTest(TestCase.java:177)
	at junit.framework.TestCase.runBare(TestCase.java:142)
	at org.eclipse.jdt.debug.tests.AbstractDebugTest.runBare(AbstractDebugTest.java:2678)
	at junit.framework.TestResult$1.protect(TestResult.java:122)
	at junit.framework.TestResult.runProtected(TestResult.java:142)
	at junit.framework.TestResult.run(TestResult.java:125)
	at junit.framework.TestCase.run(TestCase.java:130)
	at junit.framework.TestSuite.runTest(TestSuite.java:241)
	at junit.framework.TestSuite.run(TestSuite.java:236)
	at org.junit.internal.runners.JUnit38ClassRunner.run(JUnit38ClassRunner.java:90)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:93)
	at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:40)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:529)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:756)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:452)
	at org.eclipse.pde.internal.junit.runtime.RemotePluginTestRunner.main(RemotePluginTestRunner.java:74)
	at org.eclipse.pde.internal.junit.runtime.PlatformUITestHarness.lambda$0(PlatformUITestHarness.java:45)
	at org.eclipse.e4.ui.internal.workbench.swt.E4Testable.lambda$1(E4Testable.java:127)
	at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:132)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:5001)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4481)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1155)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:338)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1046)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:155)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:644)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:338)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:551)
	at org.eclipse.ui.PlatformUI.createAndRunWorkbench(PlatformUI.java:156)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:152)
	at org.eclipse.pde.internal.junit.runtime.NonUIThreadTestApplication.runApp(NonUIThreadTestApplication.java:53)
	at org.eclipse.pde.internal.junit.runtime.UITestApplication.runApp(UITestApplication.java:47)
	at org.eclipse.pde.internal.junit.runtime.NonUIThreadTestApplication.start(NonUIThreadTestApplication.java:48)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:203)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:136)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.start(EclipseAppLauncher.java:104)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:402)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:255)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:659)
	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:596)
	at org.eclipse.equinox.launcher.Main.run(Main.java:1467)
	at org.eclipse.equinox.launcher.Main.main(Main.java:1440)
```

trancexpress:
> I still saw a sporadic fail in `SuspendVMBreakpointsTests.testSuspendVmMethodExitBreakpoint()`

Can be reproduced with:

```
diff --git a/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java b/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java
index 822878216..5e897e37c 100644
--- a/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java
+++ b/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java
@@ -3363,6 +3363,7 @@ public class JDIDebugTarget extends JDIDebugElement implements
 
                @Override
                protected final IStatus run(IProgressMonitor monitor) {
+                       try { Thread.sleep(1_000); } catch (InterruptedException e1) {}
                        try {
                                runOperation(monitor);
                        } catch (DebugException e) {
```

trancexpress:
No, still fails with the last change, with "bad timing":

```
diff --git a/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java b/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java
index b35b59994..a9a141ba5 100644
--- a/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java
+++ b/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java
@@ -3368,6 +3368,7 @@ public class JDIDebugTarget extends JDIDebugElement implements
                        if (monitor.isCanceled()) {
                                return Status.CANCEL_STATUS;
                        }
+                       try { Thread.sleep(1_000); } catch (InterruptedException e1) {}
                        try {
                                runOperation(monitor);
                        } catch (DebugException e) {
```

trancexpress:
I think we need another fix idea.

iloveeclipse:
> I think we need another fix idea.

Not really. 
If I see it right, both patch and the test must be updated.
Debugging in Eclipse i see that we receive method exit event for same thread **two times**.
Patch should not schedule another suspendVM /  resumeVM operation if exact same operation for same thread is already running. I'm not sure however if that is a bug in Eclipse debugger or defined somewhere in the spec that "method exit" breakpoint is sent twice.

trancexpress:
> Not really. If I see it right, both patch and the test must be updated. Debugging in Eclipse i see that we receive method exit event for same thread **two times**. Patch should not schedule another suspendVM / resumeVM operation if exact same operation for same thread is already running. I'm not sure however if that is a bug in Eclipse debugger or defined somewhere in the spec that "method exit" breakpoint is sent twice.

And your change proposal is...?

trancexpress:
How about adding a flag to JDIThread that indicates breakpoint handling is ongoing? Then we dont touch a threas with that flag set, on resume/suspend all from JDIDebugTarget.

trancexpress:
5k test executions overnight, no deadlock.

trancexpress:
> 5k test executions overnight, no deadlock. I'll run the test some more.

One problem I see is we don't resume/suspend the thread calling the ```suspendThreads()```/```resumeThreads()```...

trancexpress:
15k test runs finished without a hang/fail.

iloveeclipse:
Whichever works :-)

trancexpress:
> Whichever works :-)

Does it actually need a rebase? I tried pulling from https://github.com/eclipse-jdt/eclipse.jdt.debug.git, I don't get more commits for ```master```.

iloveeclipse:
Looks like the build needs longer as 40 minutes now. Haven't checked what was default run time, but that sounds like a problem with the patch.

org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIThread.java
```diff
@@ -1358,6 +1363,15 @@ public synchronized IStackFrame getTopStackFrame() throws DebugException {
 	 */
 	public boolean handleSuspendForBreakpoint(JavaBreakpoint breakpoint,
 			boolean suspendVote) {
+		fHandlingSuspendForBreakpoint.set(true);
+		try {
+			return handleSuspendForBreakpointInternal(breakpoint);
+		} finally {
+			fHandlingSuspendForBreakpoint.set(true);
```
iloveeclipse, line 1370: Did you mean this:
```suggestion
			fHandlingSuspendForBreakpoint.set(false);
```

iloveeclipse (approved):
LGTM

trancexpress:
To clarify why we didn't check for suspend policy (suspend VM vs suspend only thread): retrieving the suspend policy can throw an exception, so the extra check would have to log an exception and so on. It would make the change larger and more complicated, so probably not worth the benefit of narrowing the impact of the change.

PR #74, opened 2022-06-21
Improve slow launch on big workspace (closes #73)

### LINKED ISSUE #73: Improve slow launch on big workspace
When a workspace contains a lot of libraries and a lot of projects launching can take some seconds.
image
I sampled the usecase and found that the time is used to search the path in the workspace .
In my case (~500 libraries in ~ 500 projects) 2.3 of 2.6 seconds are spend below RuntimeClasspathEntry.getResource ():
image
The searching happens because it is not known if the "path" is a workspace relative path or an absolute path to filesystem.
The problem is that searching for a filesystem path in the workspace costs number of projects times number of libraries, as seen in the call count:
image
Searching in the workspace for a relative path is much faster and should be tried first. 
That reduces the launch time in my case to almost instantly (0.3 seconds):
image

That technique was already used in Bug 546723 but did not cover the case of libraries inside the workspace.

jukzi:
getResource() for a workspace member no longer has to search for linked
resources in the whole workspace.

See issue #73 for explanation

### DIFF
--- a/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/RuntimeClasspathEntry.java
+++ b/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/RuntimeClasspathEntry.java
@@ -352,8 +352,9 @@ public class RuntimeClasspathEntry implements IRuntimeClasspathEntry {
 	protected IResource getResource(IPath path) {
 		if (path != null) {
 			IWorkspaceRoot root = ResourcesPlugin.getWorkspace().getRoot();
-			if (getType() == PROJECT) {
-				// project entry should always have a workspace relative path, try the fastest lookup first
+			if (getType() == PROJECT || path.getDevice() == null) {
+				// Project entry should always have a workspace relative path, try the fastest lookup first.
+				// Also entry without device has a great chance to find the answer quickly in the workspace.
 				IResource member = root.findMember(path);
 				if (member != null) {
 					return member;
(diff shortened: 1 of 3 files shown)


vogella:
@jjohnstn can you please review?

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/RuntimeClasspathEntry.java
```diff
@@ -352,8 +352,9 @@ public IResource getResource() {
 	protected IResource getResource(IPath path) {
 		if (path != null) {
 			IWorkspaceRoot root = ResourcesPlugin.getWorkspace().getRoot();
-			if (getType() == PROJECT) {
-				// project entry should always have a workspace relative path, try the fastest lookup first
+			if (getType() == PROJECT || path.getDevice() == null) {
+				// Project entry should always have a workspace relative path, try the fastest lookup first.
+				// Also entry without device has a great chance to find the answer quickly in the workspace.
 				IResource member = root.findMember(path);
```
SarikaSinha, line 358: What if it returns a folder? Have you tried that case?
Should we add a check for IContainer type and call the findMember again?

iloveeclipse (dismissed):

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/RuntimeClasspathEntry.java
jukzi: Also note, that org.eclipse.jdt.internal.core.JavaModel.getTarget(IPath, boolean) is used for the same purpose in for example org.eclipse.jdt.internal.core.ClasspathEntry.toString() and always checks the workspace first - independent of the type.

jukzi:
@iloveeclipse any changes do you still request? otherwise please merge

jukzi:
@iloveeclipse is there anything left? otherwise please merge

jukzi:
@SarikaSinha any changes do you still request? otherwise please merge

SarikaSinha (approved):
Looks good to me.

SarikaSinha:
@iloveeclipse 
Any objection ?

Issue #146, opened 2022-10-19
Lambda Entry breakpoint cannot be set for the selected expression


SarikaSinha:
After the fix of https://bugs.eclipse.org/bugs/show_bug.cgi?id=579100

I could not get it working even after reversal of  the change from :
https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/33

Lambda Entry breakpoint cannot be set for the selected expression, breakpoint is always set for the first lambda expression

@iloveeclipse @gayanper

gayanper:
> @gayanper
> Can you look into into this?

I can have a look at in the weekend

iloveeclipse:
@SarikaSinha : could you please provide steps to reproduce, snippet code used, used Java version etc?

SarikaSinha:
Arrays.asList("a", "b", "c").stream().filter(s -> s.length() > 0).filter(x -> x.charAt(0) > 1).filter(k -> k.indexOf(0) < 0).filter(p -> !p.isBlank()).forEach(System.out::println);

Trying to select "x ->" or "k ->" and add Lambda Entry Breakpoint, creates breakpoint for "s -> s.length() > 0" Lambda Expression.
I tried with Java 17 and 18.

gayanper:
@SarikaSinha if you have time please take this, seems like I might not be able to work on it this weekend.

gayanper:
While reproducing, for the given steps it ends up as a line breakpoint for me. But if I just keep the cursor at `filter(|x ->` then it ends up in `s ->`, thats because there is a fallback logic to find the first lambda on the line and create a lambda breakpoint there.

jukzi:
is this fixed, can be closed?

PR #147, opened 2022-10-22
Improve lambda breakpoint toggle (#146)

gayanper:
When not selecting the whole lambda expression, the codeSelect fails to find the lambda method element. So instead of using the document selection, we use the breakpoint locator location if that provides more wider range.

Signed-off-by: Gayan Perera <gayanper@gmail.com>

## Author checklist


gayanper:
@SarikaSinha even thought I couldn't reproduced as you described, I was able to see a problem in the behavior as described in the issue. Please check if this change fixes the problems you found.

SarikaSinha:
> @SarikaSinha even thought I couldn't reproduced as you described, I was able to see a problem in the behavior as described in the issue. Please check if this change fixes the problems you found.

My problem is still not resolved. I will check this week.

jukzi:
This PR need to be marked as draft, closed or finished
@SarikaSinha ping

PR #149, opened 2022-11-03
Streamline maven-toolchains-plugin used

akurtakov:
Instead of sticking to ancient version just use the version specified in eclipse-parent-pom.


rgrunber (approved):
Looks the same as https://github.com/eclipse-jdt/eclipse.jdt.core/pull/512 , except since there's only 2 usages, we don't bother defining more info in the root pom.

PR #153, opened 2022-11-25
FileHashing: avoid getCanonicalFile()

jukzi:
getCanonicalFile() is slow on windows / JDK17 - can costs even more then calculating the SHA1 of the file content, which it is supposed to avoid.

* Instead use normalized path. It does not matter that multiple files could be the same due to symbolic links, it will just recalculate their SHA1 for all of them - which is still faster.
* Also read all file attributes at once.

### DIFF
--- a/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/sourcelookup/advanced/FileHashing.java
+++ b/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/sourcelookup/advanced/FileHashing.java
@@ -17,6 +17,8 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.attribute.BasicFileAttributes;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


jukzi:
I just noticed debugging is slower on win with JDK17 then 11. root reason:
image

The code can be tested by just debugging "Hello world". Set a breakpoint in main() and run debug -> code executed.

jukzi:
@ifedorenko can you review?

jukzi:
unrelated fail: org.eclipse.jdt.debug.tests.ui.OpenFromClipboardTests.testMethod_11

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/sourcelookup/advanced/FileHashing.java
```diff
@@ -141,12 +143,17 @@ public HasherImpl(HasherImpl initial) {
...
+			if (file == null) {
 				return null;
 			}
 			try {
-				CacheKey cacheKey = new CacheKey(file);
+				BasicFileAttributes attributes = Files.readAttributes(file.toPath(), BasicFileAttributes.class);
+				if (!attributes.isRegularFile()) {
+					return null;
+				}
+				Object key= file.getAbsoluteFile().toPath().toAbsolutePath().normalize();
```
mpalat, line 155: Isn't better we check for empty path ?

mpalat (approved):
Please have a look at the review comment and incorporate the change if deemed necessary. 
+1

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/sourcelookup/advanced/FileHashing.java
jukzi: this is a java.nio.file.Path which does not have a a special case for empty. it can't be null either (checked in line 147)

jukzi:
excluded version update that was already commit by other PR

PR #174, opened 2023-01-16
Fix #91: Stack trace detection not working in colored console

mihnita:
## What it does

Removes the ANSI escape sequences before detecting links.

## How to test

See https://github.com/eclipse-platform/eclipse.platform/issues/535

The easiest way to reproduce was to create a new project and `println` the stack trace.

Attached a small file reproducing the problem.
I've tested with ANSI Support enabled and disabled, and I've checked that the links work and jump at the proper line.

## Author checklist

[Main.java.zip](https://github.com/eclipse-jdt/eclipse.jdt.debug/files/9895052/Main.java.zip)

### DIFF
--- a/org.eclipse.jdt.debug.tests/console tests/org/eclipse/jdt/debug/tests/console/JavaDebugStackTraceConsoleTest.java
+++ b/org.eclipse.jdt.debug.tests/console tests/org/eclipse/jdt/debug/tests/console/JavaDebugStackTraceConsoleTest.java
@@ -213,6 +213,31 @@ public class JavaDebugStackTraceConsoleTest extends AbstractJavaStackTraceConsol
 				new String[] { "somewhere1.somewhere2.Some1Class105", "line: 5" }, matchTexts);
 	}
 
+	public void testStackTraceDetectionWithColors() throws Exception {
+		String exception = ""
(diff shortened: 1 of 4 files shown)


mihnita:
Clean version of the closed PR #148 (identical code, work done in a feature branch instead of main)

mihnita:
Note: I don't insist on this PR, and I agree with this approach:

> I think we should not apply a quick-fix here as it unnecessary complicates things. The stacktraces are just one example and I think the ANSI console should best "hide" the ANSI nature that is when I copy/paste things, copy without escapes should be default and for API it should return a plain text and all else (HTML, RTF, with escapes) should be special cases one needs to explcitily trigger.

But it looks like I need to tinker (quite aggressively?) with `ConsoleDocument`, and in general change how things are implemented now.
So I don't think I can do it in time for the next release (2023-03).

I will 100% respect your decision if you think it is best to do this patchy fix, or live with the bug for another release
(or more? I hope not, but one never knows)

Thank you,
M.

jukzi:
Hi @mihnita, i think that PR is valid. Can you please resolve the conflict and add a junit test? See JavaDebugStackTraceConsoleTest for example.

mihnita:
Thank you!
I'll do it over the weekend.
Mihai

mihnita:
Thank you very much!

PR #181, opened 2023-01-27
Ignore 'this' by checking the name of the variable

### LINKED ISSUE #179: Missing variables when debugging Android apps
While trying to debug some Android apps in VS Code, some variables are missing from the variable list for some methods.

On more investigation, looks like it is because of the assumption that the first slot always contains this. When investigating the debug information from the compiled dex files using dexdump, looks like `this` is not always at 0th register/slot.

https://github.com/eclipse-jdt/eclipse.jdt.debug/blob/8583bcb7cce59262ca893b90d9e6ac49b5f10b0a/org.eclipse.jdt.debug/jdi/org/eclipse/jdi/internal/MethodImpl.java#L564-L566

Also, looks like it is not even required that the slots will start from 0, or that arguments will be at the beginning of the table. The logic here to check if a variable is argument or not is not valid. Also, it might already be invalid as the `argCnt` is supposed to return number of words for argument and not number of slots.

https://github.com/eclipse-jdt/eclipse.jdt.debug/blob/8583bcb7cce59262ca893b90d9e6ac49b5f10b0a/org.eclipse.jdt.debug/jdi/org/eclipse/jdi/internal/MethodImpl.java#L562

nisargjhaveri:
## What it does
Fix #179.
While getting the list of variables for a method, we were skipping the first slot assuming it was a reference to `this`. Though this is not always the case when testing with Android dex files. Instead of assuming `this` is at the first slot, check the name of the variable.

## How to test
Tested locally, the issue that was reported earlier no longer occurs.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.debug/jdi/org/eclipse/jdi/internal/MethodImpl.java
+++ b/org.eclipse.jdt.debug/jdi/org/eclipse/jdi/internal/MethodImpl.java
@@ -561,9 +561,7 @@ public class MethodImpl extends TypeComponentImpl implements Method, Locatable {
 				int slot = readInt("slot", replyData); //$NON-NLS-1$
 				boolean isArgument = slot < fArgumentSlotsCount;
 
-				// Note that for instance methods, the first slot contains the
-				// this reference.
-				if (isStatic() || slot > 0) {
+				if (!name.equals("this")) { //$NON-NLS-1$
 					LocalVariableImpl localVar = new LocalVariableImpl(
 							virtualMachineImpl(), this, codeIndex, name,
 							signature, genericSignature, length, slot,


nisargjhaveri:
@SarikaSinha, @iloveeclipse, Could you please have a look at this? Please let me know if this doesn't make sense or requires more changes.

SarikaSinha:
> @SarikaSinha, @iloveeclipse, Could you please have a look at this? Please let me know if this doesn't make sense or requires more changes.

I will be able to look at this not before next week.

SarikaSinha:
@nisargjhaveri 

Can you add some test like :
org.eclipse.debug.jdi.tests.MethodTest.testJDIVariablesByName()
org.eclipse.debug.jdi.tests.MethodTest.testJDIArguments()

nisargjhaveri:
@SarikaSinha, sure, I can add the required tests. Though, what should the new tests cover? The tests you mentioned along with `org.eclipse.debug.jdi.tests.MethodTest.testJDIVariables` seems to cover most cases I can think of.

SarikaSinha:
> @SarikaSinha, sure, I can add the required tests. Though, what should the new tests cover? The tests you mentioned along with `org.eclipse.debug.jdi.tests.MethodTest.testJDIVariables` seems to cover most cases I can think of.

It should cover the case of failure with the old code, which you have stated in the problem. And it should pass after the change ideally.

nisargjhaveri:
Okay, that way.. I believe the issue only repros in with the dex files compiled for Android. How should we go about writing tests for that? We'd need to compile to dex/APK first and then likely run Android Emulator to test?

SarikaSinha:
Yes, something like that. 
Problem in not adding the test is if someone later changes this logic, it might not work for dex files again.

nisargjhaveri:
Do we have any tests running on Android?

If not, adding the entire infra to run tests on Android would be a fairly complex work, especially for someone like me who's new to the project. While I completely agree with the fact that we should add tests for this, I don't really have enough context in this project yet. I'd be willing to take a look, but would require some more help in figuring out how to go about this.

SarikaSinha:
We don't have tests for Android. Is it possible to get the compiled classes from Android and just execute with the scenario?

nisargjhaveri:
I have attempted one POC for running tests on Android in #187. Have a look?

Android SDK provides the tool for compiling to a dex file, which runs on the dalvik vm. The Dalvik VM itself seems to be officially supported only on Android. There seems to be some ports available to run Dalvik directly on a linux machine and such, but could not find anything fully convincing. We can do some more research, if that works, makes the life much simpler.

In any case, if the current PR does not break any existing tests, can we go ahead with it and continue the discussion on the Android tests separately?

SarikaSinha:
Will release after 4.27 is complete.

nisargjhaveri:
@SarikaSinha, I believe 4.27 release should be complete now? Can we merge this now to make sure this goes into the next release?

iloveeclipse (changes requested):
Just few formal recommendations for your PR
1) Please create dedicated branch for your changes and do not commit on master branch directly.
2) Please don't merge master brsnch into your branch to get latest master state
3) Please fetch from master and **rebase** your branch on latest master state
4) Please provide commit message that explains the fix and refers to original issue https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/179

nisargjhaveri:
@iloveeclipse, Thanks for your suggestions.

Yes, I agree having a dedicated branch helps, but changing it now would likely mean I'll have to create a new PR, losing the context. And I believe it would not matter much once merged?

I've taken in the other three recommendations and updated the branch, please have a look?

iloveeclipse:
> Yes, I agree having a dedicated branch helps, but changing it now would likely mean I'll have to create a new PR, losing the context. And I believe it would not matter much once merged?

It will. Please create new PR from dedicated branch. The context is https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/179 and it will not be lost.

PR #207, opened 2023-03-22
Fix/temurin

akurtakov:
## What it does
Switches Jenkins build JVM to Temurin so latest 17.x version is used.

## How to test
Build still works

## Author checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,7 +10,7 @@ pipeline {
 	}
 	tools {
 		maven 'apache-maven-latest'
-		jdk 'openjdk-jdk17-latest'
+		jdk 'temurin-jdk17-latest'
 	}
 	stages {
 		stage('Build') {


akurtakov:
Most notable reason to use 17.0.5 is https://bugs.openjdk.org/browse/JDK-8269039 which made some jars appear as unsigned late in the 4.27 cycle. 

But even ignoring that using latest bugfix release should be the norm as these  bugfix releases are what end users end up with (JustJ https://download.eclipse.org/justj/jres/17/downloads/latest/ in EPPs) so it's best to run verification with same bugfix level.

iloveeclipse:
Do you plan / do we need to change this overall in the platform / JDT / PDE / equinox too?

akurtakov:
> Do you plan / do we need to change this overall in the platform / JDT / PDE / equinox too?

Majority are already handled. `git submodule foreach 'git grep openjdk;true'|grep Jenkins` gives 7 Jenkinsfiles. In general I handle them as I come to a repo for some reason but I would very welcome if someone wants to speed it up.

PR #211, opened 2023-03-23
Ensure that the mac os parts are installed together

laeubi:
Currently it seems if one only uses org.eclipse.jdt.launching on macos this fails with "Unbound classpath container" because the bundle org.eclipse.jdt.launching.macosx is required.

This adds a p2.inf file that adds a requirement on this at install time.

@merks @akurtakov @iloveeclipse please give this a review, I tried to adapt the SWT approach as much as possible here, I also adapted the existing p2.inf as it previously only contained a filter for x86 but not arm64, as no native code is involved here I think a simple filter for macos is sufficient.

### DIFF
--- a/org.eclipse.jdt.launching.ui.macosx/META-INF/p2.inf
+++ b/org.eclipse.jdt.launching.ui.macosx/META-INF/p2.inf
@@ -1,5 +1,4 @@
 # ensure that the applicable implementation fragment gets installed (bug 361901)
 requires.0.namespace = org.eclipse.equinox.p2.iu
 requires.0.name = org.eclipse.swt.cocoa.macosx.x86_64
-#requires.0.range = [$version$,$version$]
-requires.0.filter = (&(osgi.os=macosx)(osgi.ws=cocoa)(osgi.arch=x86_64))
+requires.0.filter = (osgi.os=macosx)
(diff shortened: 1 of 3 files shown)


merks:
Might this problem be better fixed in PDE when it computes the required bundles such that it generally considers a fragment required if its host is required and the fragment filters match?  I.e., I expect this is a general problem, which is why I mostly try to launch based on features rather than bundles.  I wonder, for example, why this same problem doesn't arise also for org.eclipse.core.resource's and org.eclipse.core.filesystem's fragment.

In any case, I don't have a problem with adding such a thing for JDT, though I see it currently breaks the build...

laeubi:
This is not a problem with PDE but if one requests P2 to install a product or e.g. run a Plugin-Test with Tycho, you can see that m2e currently has a workaround for this here:
https://github.com/eclipse-m2e/m2e-core/blob/270e3569672b5a31572dc671f625bcc580987828/pom.xml#L255-L263

Also please note that this is not a **fragment** but a **bundle** even though I'm not sure why this was implemented by a separate one, I think maybe so one don't need to decide on runtime.

laeubi:
Closing this as it seems of no interest for JDT to improve the situation using p2.inf.

merks:
This looks almost just like what SWT is doing:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/1190e3c6ca6b4de5817e29863d1e0a683a1a6e96/bundles/org.eclipse.swt/pom.xml#L48-L55

laeubi:
Now we have
- https://github.com/eclipse-tycho/tycho/pull/3263

there is a good chance this works now!

laeubi:
> In any case, I don't have a problem with adding such a thing for JDT, though I see it currently breaks the build...

Build is working now, not sure how the test-failure might be related?

akurtakov:
Thanks @laeubi !

Issue #217, opened 2023-03-27
org.eclipse.jdt.debug.ui/snippetsupport.jar, if rebuilt, breaks snippet editor


iloveeclipse:
As title says, in context of https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/210 I've seen that org.eclipse.jdt.debug.ui/snippetsupport.jar should be rebuilt because related code changed.

So I did.

... And the new jar simply broke JavaSnippetEditor (`Ctrl+3 -> create Java Scrapbook`): the editor can be opened, editing works but no execution / evaluation possible.

I've first assumed my changes were wrong etc, but in fact, recompilation of the `snippetsupport.jar`  (which can be triggering by touching `org.eclipse.jdt.debug.ui/scripts/buildExtraJAR.xml` in the IDE) breaks `JavaSnippetEditor` on master too. 

Actually it doesn't work since 205c5a0c1fcd1d7b72a0af85c79632f7b522d90a, more precisely, [since this change](https://github.com/eclipse-jdt/eclipse.jdt.debug/blame/205c5a0c1fcd1d7b72a0af85c79632f7b522d90a/org.eclipse.jdt.debug.ui/Snippet%20Support/org/eclipse/jdt/internal/debug/ui/snippeteditor/ScrapbookMain1.java#LL4C11-L4C11).

The problem is coming from this line: https://github.com/eclipse-jdt/eclipse.jdt.debug/blob/205c5a0c1fcd1d7b72a0af85c79632f7b522d90a/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/snippeteditor/JavaSnippetEditor.java#L1117

where the debugger expects a frame with dedicated line number of the **last line** in the  `ScrapbookMain1.eval()` method...
Well, after updating license info the line changed, condition doesn't match and evaluation doesn't works anymore.

This was undetected because there was no rebuild of the org.eclipse.jdt.debug.ui/snippetsupport.jar **since 8 years**!
Last time this file was generated & pushed to git via 051fce25278ddaccf5c402484bf105ce2a390160.

I plan to provide a patch for this issue.

PR #223, opened 2023-03-28
Fix for "snippetsupport.jar, if rebuilt, breaks snippet editor" (#217)

iloveeclipse:
The actual fix consists in `==` replaced with `>=` in `lineNumber >= SCRAPBOOK_MAIN1_LAST_LINE`, so we match also if the line moved down (as already happened), the rest is just a bit of (necessary) cleanup.

The fix will not work if the line will be moved up, but for that case we have the `JavaSnippetEditorTest.testEvaluation()` test that should catch the (breaking) inconsistency between `JavaSnippetEditor` and the compiled `ScrapbookMain1` class.

Note: `SCRAPBOOK_MAIN1_LAST_LINE` is still pointing to line 28 (and not 31) to avoid unneeded extra re-compilation / checkin of `org.eclipse.jdt.debug.ui/snippetsupport.jar`
which is (as of today) checked into git and NOT rebuilt automatically.

The content of `snippetsupport.jar` was built and checked into git last time via 051fce25278ddaccf5c402484bf105ce2a390160 and doesn't match *current* content of ScrapbookMain1 that was updated later via 205c5a0c1fcd1d7b72a0af85c79632f7b522d90a without rebuilding the jar file.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/217

### DIFF
--- a/org.eclipse.jdt.debug.ui/Snippet Support/org/eclipse/jdt/internal/debug/ui/snippeteditor/ScrapbookMain1.java
+++ b/org.eclipse.jdt.debug.ui/Snippet Support/org/eclipse/jdt/internal/debug/ui/snippeteditor/ScrapbookMain1.java
@@ -29,5 +29,13 @@ public class ScrapbookMain1 {
 	public static void eval(Class<?> clazz) throws ClassNotFoundException, NoSuchMethodException, InvocationTargetException, IllegalAccessException {
 		Method method=clazz.getDeclaredMethod("nop", new Class[0]); //$NON-NLS-1$
 		method.invoke(null, new Object[0]);
+		// XXX: if changing this class, make sure to check if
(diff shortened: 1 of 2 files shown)


iloveeclipse:
Unrelated fail:
```
java.lang.Error: Target program was not ready after 10000ms
	at org.eclipse.debug.jdi.tests.AbstractJDITest.waitUntilReady(AbstractJDITest.java:1544)
	at org.eclipse.debug.jdi.tests.AbstractJDITest.triggerAndWait(AbstractJDITest.java:1406)
	at org.eclipse.debug.jdi.tests.AbstractJDITest.triggerAndWait(AbstractJDITest.java:1364)
	at org.eclipse.debug.jdi.tests.MethodEntryRequestTest.testJDIWithClassFilter2(MethodEntryRequestTest.java:125)
```

Issue #230, opened 2023-04-04
Detect Java installations from standard folders


mickaelistria:
The use-case is the following:
I have Java 19 installed in standard `/usr/lib/jvm/...` directory and am opening Eclipse IDE hoping to work with Java 19. Then Java 19 doesn't work by default, no compatible runtime is found, classpath entry is unbound.
JDT debug could have some pre-defined standard directories to detect Java installations from; according to the OS, similarly to how it is done for web browsers detection.

rgrunber:
In vscode-java we use an npm library jdk-utils mainly to try to detect a compatible (Java 17) JRE if we're on a platform where we don't ship our own. https://github.com/Eskibear/node-jdk-utils/blob/main/src/index.ts#L81 . Are there any locations in that logic that might be beneficial to add here also ?

@testforstephen, @fbricon this should be a no-op improvement on our end right ? When build support for m2e/gradle detect the source/target, it seems to be matched with the default JDK (eg. 17) (where we don't set `java.configuration.runtimes`). Might this change handle setting it correctly ?

fbricon:
@rgrunber yes jdt.ls will get it for free. 

vscode-java could also ship https://github.com/sidespin/jre-discovery which is a more complete JRE detection solution (extensible, doesn't require restart)

basilevs:
Does it make sense to add `${JAVA_HOME}/..` to root directories?

mickaelistria:
> Does it make sense to add ${JAVA_HOME}/.. to root directories?

Yes, I think so. As I'm now mostly using JustJ, I find the running VM is often missing.

PR #231, opened 2023-04-04
Detect JVM installs at startup

mickaelistria:
Fixes https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/230

## What it does

This adds a (opt-out) startup job that looks up installed JRE is some standard directores. So people usually don't need to configure JRE by hand. It's useful both to Eclipse IDE and JDT-LS with VSCode.

=## How to test

Start a fresh workspace, go to Installed JREs preference page, enjoy that some JREs are detected without any effort.

## Author checklist


org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/LaunchingPlugin.java
```diff
@@ -582,6 +589,66 @@ public void saving(ISaveContext context1) throws CoreException {
 		ResourcesPlugin.getWorkspace().addResourceChangeListener(this, IResourceChangeEvent.PRE_DELETE | IResourceChangeEvent.PRE_CLOSE);
 		DebugPlugin.getDefault().getLaunchManager().addLaunchListener(this);
 		DebugPlugin.getDefault().addDebugEventListener(this);
+		IEclipsePreferences instanceNode = InstanceScope.INSTANCE.getNode(getBundle().getSymbolicName());
+		IEclipsePreferences defaultNode = DefaultScope.INSTANCE.getNode(getBundle().getSymbolicName());
+		if (instanceNode.getBoolean(PREF_DETECT_VMS_AT_STARTUP, defaultNode.getBoolean(PREF_DETECT_VMS_AT_STARTUP, true))) {
+			new Job(LaunchingMessages.lookupInstalledJVMs) {
```
trancexpress, line 595: Can code from this job be extracted in a separate class?

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/LaunchingPlugin.java
```diff
@@ -582,6 +589,66 @@ public void saving(ISaveContext context1) throws CoreException {
...
+				@Override
+				protected IStatus run(IProgressMonitor monitor) {
+					Collection<File> rootDirectories = new HashSet<>();
+					rootDirectories.add(new File("/usr/lib/jvm")); //$NON-NLS-1$
```
trancexpress, line 599: I assume this will do nothing for Windows, how does the preference page address this (I didn't see if there is some switch which doesn't show the new preference UIs / didn't see anything in the message to indicate no Windows support)?

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/LaunchingPlugin.java
```diff
@@ -582,6 +589,66 @@ public void saving(ISaveContext context1) throws CoreException {
 		ResourcesPlugin.getWorkspace().addResourceChangeListener(this, IResourceChangeEvent.PRE_DELETE | IResourceChangeEvent.PRE_CLOSE);
 		DebugPlugin.getDefault().getLaunchManager().addLaunchListener(this);
 		DebugPlugin.getDefault().addDebugEventListener(this);
+		IEclipsePreferences instanceNode = InstanceScope.INSTANCE.getNode(getBundle().getSymbolicName());
```
iloveeclipse, line 592: As usually, please check if that works with product customization (e.g. disable automatic search)

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/LaunchingPlugin.java
```diff
@@ -582,6 +589,66 @@ public void saving(ISaveContext context1) throws CoreException {
...
+		IEclipsePreferences defaultNode = DefaultScope.INSTANCE.getNode(getBundle().getSymbolicName());
+		if (instanceNode.getBoolean(PREF_DETECT_VMS_AT_STARTUP, defaultNode.getBoolean(PREF_DETECT_VMS_AT_STARTUP, true))) {
+			new Job(LaunchingMessages.lookupInstalledJVMs) {
+				@Override
+				protected IStatus run(IProgressMonitor monitor) {
```
trancexpress, line 597: Considering there will be I/O here, might make sense to make use of the progress monitor and check for cancellation.

iloveeclipse (changes requested):

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/LaunchingPlugin.java
mickaelistria: Yes, I'll do that tomorrow.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/LaunchingPlugin.java
mickaelistria: This particular directory will be ignored. I think the `.sdkman` folder is also used on Windows; if so the job can detect such installations on Windows.
However, if you can suggest some default directories to look at for Windows, I'd be glad to add them to this list. I just happen to not know how a JDK in installed usually on Windows.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/LaunchingPlugin.java
mickaelistria: Ack.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/LaunchingPlugin.java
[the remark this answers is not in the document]
mickaelistria: Ack, I'll take care of it tomorrow.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/LaunchingPlugin.java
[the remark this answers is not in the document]
mickaelistria: As far as I know of SDKMan (ie not much), the `.sdkman` makes sense on Windows.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/LaunchingPlugin.java
mickaelistria: Verified.

mickaelistria:
All tests are now passing. I had to adjust some of them because those were assuming that no other JRE is configured for proper test executions; which is now wrong on CI as the infra has numerous JVMs that are now detected, some of them without sources. Adding extra checks on the tests make them work more reliably on any infra.

SarikaSinha:
@mickaelistria 
I will look at the code in the next few days. Had some basic queries :
1. Does it handle only to populate the JREs in Installed JREs block or also uses them to launch if there was none defined in ini or in the folder under the eclipse directory?
2. It does not break anything defined here https://wiki.eclipse.org/Eclipse.ini ?

mickaelistria:
> Does it handle only to populate the JREs in Installed JREs block

Yes.

> also uses them to launch if there was none defined in ini or in the folder under the eclipse directory?

It adds the VMs so JDT known them, then JDT may then use it or not in order to run particular programs or resolve execution environments. I did not touch the code that is responsible for choosing the VM to use by default when computing a project build path or when running it. So if it's hardcoded somewhere than the one in eclipse.ini should have priority, it should still happen.

> It does not break anything defined here https://wiki.eclipse.org/Eclipse.ini ?

The patch doesn't affect the Eclipse launch in any way. So what is defined in eclipse.ini and how it is used in order to start the Eclipse product is not at all modified by this patch.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/jres/JREsPreferencePage.java
```diff
@@ -167,6 +174,9 @@ public void selectionChanged(SelectionChangedEvent event) {
 				}
 			}
 		});
+
+		SWTFactory.createVerticalSpacer(ancestor, 1);
```
iloveeclipse, line 178: This preference is hardly detectable at the bottom of the table, it should be *above* the table with available JRE's

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/jres/JREsPreferencePage.java
```diff
@@ -87,6 +92,8 @@ public JREsPreferencePage() {
 	 */
 	@Override
 	public void init(IWorkbench workbench) {
+		setPreferenceStore(new ScopedPreferenceStore(InstanceScope.INSTANCE,
```
iloveeclipse, line 95: Haven't debugged that, but why do we need to set a preference store for the **existing** preference page? How did it worked before?

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
```diff
@@ -0,0 +1,121 @@
...
+		super(LaunchingMessages.lookupInstalledJVMs);
+	}
+
+	@Override
+	protected IStatus run(IProgressMonitor monitor) {
+		Collection<File> rootDirectories = new HashSet<>();
+		if (!Platform.OS_WIN32.equals(Platform.getOS())) {
+			rootDirectories.add(new File("/usr/lib/jvm")); //$NON-NLS-1$
+		}
+		rootDirectories.add(new File(System.getProperty("user.home"), ".sdkman/candidates/java")); //$NON-NLS-1$ //$NON-NLS-2$
```
iloveeclipse, line 48: Shouldn't that also go under "non-Windows" if block? `.sdkman/candidates/java`

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
```diff
@@ -0,0 +1,121 @@
...
+		candidates.removeIf(knownInstallations::contains);
+		StandardVMType standardType = (StandardVMType)JavaRuntime.getVMInstallType(StandardVMType.ID_STANDARD_VM_TYPE);
+		candidates.forEach(f -> {
+			if (monitor.isCanceled()) {
+				return;
+			}
+			new Job(NLS.bind(LaunchingMessages.configuringJVM, f.getAbsolutePath())) {
+				@Override
+				public IStatus run(IProgressMonitor monitor) {
+					VMStandin workingCopy = new VMStandin(standardType, f.getAbsolutePath());
```
iloveeclipse, line 96: Somewhere along this code a check is missing that skips non-valid "JRE" directories which aren't actually JRE's. Just export  /tmp in JDK_HOME and enjoy various crazy "JRE" directories found.

After next startup they disappear but surely we won't confuse anyone in case some garbage input is accepted here as "JRE".

I believe `org.eclipse.jdt.internal.launching.StandardVMType.validateInstallLocation(File)` is needed.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
```diff
@@ -0,0 +1,121 @@
...
+					workingCopy.setInstallLocation(f);
+					workingCopy.setName(f.getName());
+					workingCopy.convertToRealVM();
```
iloveeclipse, line 99: Not sure how "multi-threaded" the code here is, as it probably was written considering one single user action is doing all that.
Do the N jobs need to run in parallel, or can we just do that work here in the current job, without extra jobs?
Just report as sub-tasks for monitor and loop over candidates?

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
```diff
@@ -0,0 +1,121 @@
...
+				public boolean belongsTo(Object family) {
+					return family.equals(FAMILY);
+				}
+
+			}.schedule();
+		});
+		if (monitor.isCanceled()) {
+			return Status.CANCEL_STATUS;
+		}
+		return Status.OK_STATUS;
```
iloveeclipse, line 113: I wonder if we should automatically set the property to "false" at the end of the job.
I don't think we want add extra work discovering / validating all the VM's on every startup given the fact that in most cases all the work will be useless because all the VM's are already detected / validated on first workspace start.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/LaunchingPlugin.java
```diff
@@ -582,6 +584,11 @@ public void saving(ISaveContext context1) throws CoreException {
 		ResourcesPlugin.getWorkspace().addResourceChangeListener(this, IResourceChangeEvent.PRE_DELETE | IResourceChangeEvent.PRE_CLOSE);
 		DebugPlugin.getDefault().getLaunchManager().addLaunchListener(this);
 		DebugPlugin.getDefault().addDebugEventListener(this);
+		IEclipsePreferences instanceNode = InstanceScope.INSTANCE.getNode(getBundle().getSymbolicName());
+		IEclipsePreferences defaultNode = DefaultScope.INSTANCE.getNode(getBundle().getSymbolicName());
+		if (instanceNode.getBoolean(PREF_DETECT_VMS_AT_STARTUP, defaultNode.getBoolean(PREF_DETECT_VMS_AT_STARTUP, true))) {
```
iloveeclipse, line 589: Could you please extract `defaultNode.getBoolean(PREF_DETECT_VMS_AT_STARTUP, true)` to a variable, to simplify reading / debugging?

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/LaunchingPlugin.java
```diff
@@ -582,6 +584,11 @@ public void saving(ISaveContext context1) throws CoreException {
...
+			new DetectVMInstallationsJob().schedule(200);
```
iloveeclipse, line 590: Delay of 200 is needed because of ... ? Do we need it at all?

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
```diff
@@ -0,0 +1,121 @@
...
+class DetectVMInstallationsJob extends Job {
+
+	private static final Object FAMILY = "Family " + DetectVMInstallationsJob.class.getName(); //$NON-NLS-1$
+	
+	DetectVMInstallationsJob() {
+		super(LaunchingMessages.lookupInstalledJVMs);
+	}
+
+	@Override
+	protected IStatus run(IProgressMonitor monitor) {
```
iloveeclipse, line 43: The method is long.
Can we please split it into:
1) get known VM's
2) get candidate VM's
3) create new VM's for those from 2) who are not in 1)

It would be easier to read.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
```diff
@@ -0,0 +1,121 @@
...
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.Job;
+import org.eclipse.jdt.launching.IVMInstall;
+import org.eclipse.jdt.launching.IVMInstallType;
+import org.eclipse.jdt.launching.JavaRuntime;
+import org.eclipse.jdt.launching.VMStandin;
+import org.eclipse.osgi.util.NLS;
+
```
iloveeclipse, line 33: Please add some basic javadoc what is the purpose of this job.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
```diff
@@ -0,0 +1,121 @@
...
+class DetectVMInstallationsJob extends Job {
+
+	private static final Object FAMILY = "Family " + DetectVMInstallationsJob.class.getName(); //$NON-NLS-1$
```
iloveeclipse, line 36: I would prefer either `DetectVMInstallationsJob.class` or `DetectVMInstallationsJob.class.getName()`.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
```diff
@@ -0,0 +1,121 @@
...
+	
+	DetectVMInstallationsJob() {
+		super(LaunchingMessages.lookupInstalledJVMs);
+	}
+
+	@Override
+	protected IStatus run(IProgressMonitor monitor) {
+		Collection<File> rootDirectories = new HashSet<>();
+		if (!Platform.OS_WIN32.equals(Platform.getOS())) {
```
iloveeclipse, line 45: Please check `org.eclipse.jdt.internal.debug.ui.jres.InstalledJREsBlock.search()` for special Mac code. I have no idea about Macs, but may be you would be inspired by that.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
```diff
@@ -0,0 +1,121 @@
...
+		candidates.forEach(f -> {
+			if (monitor.isCanceled()) {
+				return;
+			}
+			new Job(NLS.bind(LaunchingMessages.configuringJVM, f.getAbsolutePath())) {
+				@Override
+				public IStatus run(IProgressMonitor monitor) {
+					VMStandin workingCopy = new VMStandin(standardType, f.getAbsolutePath());
+					workingCopy.setInstallLocation(f);
+					workingCopy.setName(f.getName());
```
iloveeclipse, line 98: `f.getName()` is surely not unique, so if any of candidates located at different root directories have same names, only the last (or the first?) will survive, others will cause errors or be ignored. I haven't checked where they are "added" to the list of "remembered" JRE's, but at least on the UI it is not possible to add new JRE that has same name as one of already defined.

iloveeclipse (changes requested):

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
mickaelistria: I think SDKMan uses this directory on Windows as well.

iloveeclipse: But surely not as `/candidates/java` ? Or will Java convert that path to `\candidates\java` on Windows automatically?

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/ClasspathVariableTests.java
```diff
@@ -65,7 +65,6 @@ public void testJRELibResolution() throws CoreException {
 		assertNotNull("no default JRE", vm);
 		LibraryLocation[] libs = JavaRuntime.getLibraryLocations(vm);
 		assertTrue("no default libs", libs.length > 0);
-		assertEquals("Should resolve to location of local JRE", libs[0].getSystemLibraryPath().toOSString().toLowerCase(), resolved[0].getPath().toOSString().toLowerCase());
```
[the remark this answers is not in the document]
mickaelistria: What "resolved variables" do you have in mind. The issue here is that this test is actually wrong and does not test what the code actually implements. There is no reason the default VM is used here; it just happened to work on CI for a while because only 1 JRE was known to JDT.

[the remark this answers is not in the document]
iloveeclipse: `resolved` variable above should be in the list of `libs`

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/jres/JREsPreferencePage.java
mickaelistria: The preference page uses other JDT APIs to interact with the configuration, see performOK. Setting the store allows `super.performOK` to store this new preference as usual.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/jres/JREsPreferencePage.java
mickaelistria: Ack.

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/sourcelookup/Bug565462Tests.java
```diff
@@ -78,8 +87,20 @@ public void testFindDuplicatesBug565462() throws Exception {
...
+		assertEquals("Expected only 1 match for class " + className + " in " + JavaRuntime.computeVMInstall(configuration).getInstallLocation() + " but found: " + Arrays.toString(foundElements), 1, foundElements.length);
+
+	}
+
+	private static void ensureHasSource(IExecutionEnvironment ee) {
+		Predicate<IVMInstall> hasSource = vm -> vm != null && vm.getInstallLocation() != null && new File(vm.getInstallLocation(), "lib/src.zip").isFile();
+		if (!hasSource.test(ee.getDefaultVM())) {
+			Stream.of(ee.getCompatibleVMs()).filter(Objects::nonNull).filter(hasSource)
+				.findAny()
+				.ifPresent(ee::setDefaultVM);
```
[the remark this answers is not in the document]
mickaelistria: Ack.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
mickaelistria: Ack

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
mickaelistria: Checking for existing VMs is cheap. And one of the goals with this change is to get updates to newer Java (which can be more frequent than bootstrapping an Eclipse workspace) automatically surfacing in the IDE so users can leverage those new installations with less configuration.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/LaunchingPlugin.java
mickaelistria: Ack

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/LaunchingPlugin.java
mickaelistria: Ack

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
mickaelistria: Ack

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
mickaelistria: Ack

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
mickaelistria: Ack

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
mickaelistria: OK. Should we use the absolute path as name then?

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/ClasspathVariableTests.java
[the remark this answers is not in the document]
mickaelistria: Unlike what the test has assumed,, the "libs" does *not* relate to the default VM; but to the VM that JDT think is best for the project. So the libs doesn't contain the (incorrectly) previously expected value. I don't think such an assert should and can be made to work in general.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
mickaelistria: I believe that either Java or Windows convert to `\` automatically (in `FileSystem.normalize()` I believe). If you have Windows, you can quickly test with `jshell`.
However, if you think it's important, I don't mind replacing `/` with `File.separator`

mickaelistria: https://stackoverflow.com/questions/2417485/difference-between-file-separator-and-slash-in-paths contains answers with references that claim `/` is OK to use in File on Windows as well.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
mickaelistria: I verified and the check in UI is not an API requirement. I could install multiple JVMs with same name
Screenshot from 2023-04-17 17-02-19
 (via API with this code, copying JVM installations to JDK_HOME)

mickaelistria: OK, I've changed it and copied the logic from the "Search" button to suffix with `(1)`, `(2)`,...

mickaelistria:
@iloveeclipse Any remaining issues you think need to be considered before we merge?
@SarikaSinha IIRC, you're using a Mac, aren't you? If so, and if you feel like it, feel free to enrich this PR with whatever makes it work better on mac.

mickaelistria:
> Indeed, failing test :-)

Thanks for the reminder, I had forgotten about it and will try to fix it shortly.

iloveeclipse:
Would be great if you could run SDK JDT core / ui / PDE tests locally because I fear they might be affected as well by the change.

If you can't - **note** for whoever will merge that - please run an extra SDK build so in case we will see multiple related test fails, we could quickly revert the change, so no one would be blocked by failing tests.

But after writing that I realised that we most likely should disable the automated JDK detection job for tests via maven system property.

What I fear are two things:
1) Random time where the job may run and affect the system while tests are running (and probably would be affected by the job code doing something in JDT in parallel). We have already enough troubles with unstable tests, we don't want make it worse. 
2) The fact that we may see **multiple** system dependent JDK's found by Eclipse and so changing test environment that expected to see only one JVM configured.

mickaelistria:
> But after writing that I realised that we most likely should disable the automated JDK detection job for tests via maven system property.

OK.

SarikaSinha:
> @iloveeclipse Any remaining issues you think need to be considered before we merge? @SarikaSinha IIRC, you're using a Mac, aren't you? If so, and if you feel like it, feel free to enrich this PR with whatever makes it work better on mac.

Yes, I am using a Mac. Will check out.

iloveeclipse:
@mickaelistria : may be I''ve missed that: but have you addressed [my concerns above](https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/231#issuecomment-1513171084)? Is it possible to disable job startup for tests via a system property?

iloveeclipse:
@SarikaSinha : why did you "merge" master to this branch? Please use *rebase*.

SarikaSinha:
> @SarikaSinha : why did you "merge" master to this branch? Please use _rebase_.

I used Update Branch from the UI, I thought it rebases ?

SarikaSinha:
> > I used Update Branch from the UI, I thought it rebases ?
> 
> Nope: image

Can we change the default to "Update with Rebase" ?

iloveeclipse:
> Can we change the default to "Update with Rebase" ?

No, at least I'm unaware about such option.

SarikaSinha:
> > Can we change the default to "Update with Rebase" ?
> 
> No, at least I'm unaware about such option.

It's not there yet, users are requesting.

SarikaSinha:
@mickaelistria 
For Macos, we can add "/Library/Java/JavaVirtualMachines"  to the root directory.

mickaelistria:
Latest patch attempts to address all remaining concerns.

mickaelistria:
Anything I can do to facilitate progress here?

mickaelistria:
> Should we ask for help @jjohnstn or @rgrunber in this case?

Yes, that would probably be helpful.

akurtakov:
@iloveeclipse Are you fine with @jjohnstn taking over?

iloveeclipse:
> @iloveeclipse Are you fine with @jjohnstn taking over?

No objections. However, I would prefer to postpone this and plan for 4.29 M1.

mickaelistria:
Hi,
Can this be considered for a review/merge soon, so we get time to gather feedback during 4.29 cycle and adapt whatever needs to be adapted in a comfortable timeframe?

akurtakov:
@jjohnstn Would you please review this one?

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
jjohnstn: Check out: https://stackoverflow.com/questions/15826202/where-is-java-installed-on-mac-os-x
for description of logic.  It appears that on Mac, JRE installs are not put in the JavaVirtualMachines folder specified later in this file and you need to use /usr/libexec/java_home.   That said, why not simply change the code on MacOS to call: MacInstalledJREs.getInstalledJREs() which is static, uses the proper logic, and does all the work for you to create an array of VMStandin?

jjohnstn (changes requested):

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
mickaelistria: I really cannot properly do the job for mac. There are too many things I don't know and if I do it wrong, it causes more bugs that it fixes issues.
For example: does MacInstalledJREs add standins for already known VMs? Am I supposed to filter the lists? Are the installations referencing the Java installation home or some subfolder that would still be valid? Do I have to verify the name is unique or is it already done...
There are too many things I'm not comfortable in implementing just by guess; I would rather see this patch merged as it, or even disabled on Mac, and let someone with good knowledge of mac integration and ability to test fill the holes.

jjohnstn: Open a new issue for Detect JVM installs at startup on Mac.  I'll look at it for M2.

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/ClasspathVariableTests.java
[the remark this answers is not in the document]
jjohnstn: I can confirm that running the test locally that the project is only requiring 1.4 and it is satisfied by my local 1.8 but the default VM is Java 17 so the test fails if I try and find a match between libs and resolved.  Marking this removal as ok.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
mickaelistria: Done with https://github.com/redhat-developer/vscode-java/pull/3152

jjohnstn (approved):

jjohnstn:
@mickaelistria Please add a N&N entry.  Patch approved.

iloveeclipse:
Note: we see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1150 today, may be caused by this PR.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
```diff
@@ -0,0 +1,154 @@
...
+		Set<File> directories = rootDirectories.stream().filter(File::isDirectory)
+			.map(dir -> dir.listFiles(File::isDirectory))
+			.flatMap(Arrays::stream)
+			.filter(Objects::nonNull)
+			.collect(Collectors.toSet());
+
+		// particular VM installations
+		String javaHome = System.getenv("JAVA_HOME"); //$NON-NLS-1$
+		if (javaHome != null) {
+			directories.add(new File(javaHome));
```
basilevs, line 111: Directories may be unmodifiable. See [Collectors.toSet()](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/stream/Collectors.html#toSet())
> There are no guarantees on the type, mutability, serializability, or thread-safety of the Set returned; if more control over the returned Set is required, use [toCollection(Supplier)](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/stream/Collectors.html#toCollection(java.util.function.Supplier)).

PR #242, opened 2023-04-21
Delete classpath argument files upon shutdown. Fixes #240

### LINKED ISSUE #240: "Show Commandline" leaks vmargs file
This is a copy of https://github.com/eclipse-platform/eclipse.platform.debug/issues/84
Summary: Open "Show Commandline" for a project's launch configuration with a large class/module path; this results in a file containing the classpath, which is then used as parameter like @/home/.../.temp-Main--module-path-arg-1681976925916.txt. After closing the dialog this file remains.

A solution could be to store these files in another location, not in the workspace. Any other approach will most likely result in interface changes.

diti0023:
View also [original issue](https://github.com/eclipse-platform/eclipse.platform.debug/issues/84) for more information.

## What it does
In the dialog Run Configurations, click on "Show Comandline". If there are many or long classpaths, they are collected in a file in order not to exceed the maximum command length (on a Linux system check with `getconf ARG_MAX`). This is also done for actually launching the run, but in that case these files are removed on terminating the launch. When just showing the command line these files remain, in order to have them accessible later (i.e. after closing the dialog). In this case, i.e. without any process, the GUI has no access to these files. This change marks the files as to be deleted on shutdown of the JVM, and handles them via a .gitignore in order to prevent accidentally committing them.

## How to test
Create a project with many JAR files and long paths to them. When the command length of the OS is about to be exceeded, the resulting command changes from listing all files as parameters (-classpath) to a single entry like

`@/home/.../.temp-Main--module-path-arg-1681985065801.txt`

The files resulting from just showing the commandline are deleted when the JVM is terminated.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/ClasspathShortener.java
+++ b/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/ClasspathShortener.java
@@ -312,7 +312,6 @@ public class ClasspathShortener implements IProcessTempFileCreator {
 			Files.writeString(argFile.toPath(), arg, systemCharset);
 			argFile.deleteOnExit();
 			file = argFile;
-			// createOrUpdateGitIgnore();
 		} catch (IOException e) {
 			throw new CoreException(new Status(IStatus.ERROR, LaunchingPlugin.getUniqueIdentifier(), IStatus.ERROR, "Cannot create " + option //$NON-NLS-1$
 					+ " argument file", e)); //$NON-NLS-1$
--- a/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/ClasspathShortener.java
+++ b/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/ClasspathShortener.java
@@ -32,6 +32,7 @@ import java.util.jar.Manifest;
 
 import org.eclipse.core.runtime.Assert;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.Status;
@@ -66,6 +67,7 @@ public class ClasspathShortener implements IProcessTempFileCreator {
 	private String[] envp;
 	private File processTempFilesDir;
 	private final List<File> processTempFiles = new ArrayList<>();
+	private static final String ARGFILE_TEMPDIR_NAME = "temp"; //$NON-NLS-1$
 
 	/**
 	 *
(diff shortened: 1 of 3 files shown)


org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/ClasspathShortener.java
```diff
@@ -302,6 +302,7 @@ private void shortenPathUsingClasspathArgumentFile(int modulePathArgumentIndex,
 						+ systemCharset.displayName() + ".", null)); //$NON-NLS-1$
 			}
 			Files.writeString(argFile.toPath(), arg, systemCharset);
+			argFile.deleteOnExit();
```
jukzi, line 305: While this is an improvement it would not solve the issue that developers complain that the file shows up in the git staging view while they have eclipse open. So they could still accidentally commit it.

diti0023: That's true, and this is why this is still a draft ...

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/ClasspathShortener.java
```diff
@@ -313,6 +317,25 @@ private void shortenPathUsingClasspathArgumentFile(int modulePathArgumentIndex,
 		addProcessTempFile(file);
 	}
 
+	private void createOrUpdateGitIgnore() {
+		File gitIgnore = new File(processTempFilesDir, ".gitignore"); //$NON-NLS-1$
```
iloveeclipse, line 321: Instead of touching gitignore, one can contribute something like:
```
  <extension
        point="org.eclipse.team.core.ignore">
     <ignore
           enabled="true"
           pattern=".temp-*path-arg-*.txt"/>
  </extension>
```

iloveeclipse:
> [Merge branch 'eclipse-jdt:master' into issue#240](https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/242/commits/a3d095c62a4a23147cf4868c800dd2db20d7b61f)

@diti0023 : please don't use merge to get latest state from master, please use **rebase**.
Merge commits are not allowed to get clean git history.

diti0023:
> > [Merge branch 'eclipse-jdt:master' into issue#240](https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/242/commits/a3d095c62a4a23147cf4868c800dd2db20d7b61f)
> 
> @diti0023 : please don't use merge to get latest state from master, please use **rebase**. Merge commits are not allowed to get clean git history.

@iloveeclipse  Thanks -- obviously I'm still a newbie ...

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/ClasspathShortener.java
diti0023: Unfortunately neither ".temp-*path-arg-*.txt" nor "**/.temp-*path-arg-*.txt" actually hides these files in the Package Explorer; I guess I have to investigate further.

iloveeclipse: > would it be feasible to use a dedicated temp path below
workspace\.metadata\.plugins\org.eclipse.jdt.launching?

Technically we can create these temp files there, yes. I wonder why that wasn't done and if there is a catch (like having them in the base directory etc).

diti0023:
Current solution:
For "Show command line" the temp files are stored in the metadata of the respective plug-in, so outside the workspace. Additionally the files are deleted at shutdown of the Eclipse instance.

iloveeclipse:
Note: we are in 4.28 RC phase now and I personally would prefer that we would postpone this PR to 4.29 M1.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/ClasspathShortener.java
```diff
@@ -302,7 +310,9 @@ private void shortenPathUsingClasspathArgumentFile(int modulePathArgumentIndex,
 						+ systemCharset.displayName() + ".", null)); //$NON-NLS-1$
 			}
 			Files.writeString(argFile.toPath(), arg, systemCharset);
+			argFile.deleteOnExit();
 			file = argFile;
+			// createOrUpdateGitIgnore();
```
iloveeclipse, line 315: Can you please remove this comment?

diti0023: Sure, just overlooked this.

PR #246, opened 2023-05-24
Set console stream encoding for java >= 19 - eclipse.platform.debug#124

jukzi:
## What it does
Set "-Dstdout.encoding" and "-Dstderr.encoding" to the value specified in launch configration:
to fix https://github.com/eclipse-platform/eclipse.platform/issues/530
This does NOT fix the issue for JDK 18, as there is no official parameter for that JVM.
Relevant is the Java Version of the JVM to launch.

## How to test
Launch  a programm that outputs non-Ascii chars. For example:
image

```
public class ConsoleTest {
		System.out.println("hällö€");
		System.err.println("hällö€");
	}
}
```
Tested on JDK 17, JDK 19 with UTF-8, ISO-8859-1, windows-1252
image
image

## Author checklist


jukzi:
unrelated random fail https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/258

PR #259, opened 2023-06-09
JavaSnippetEditor: fix missing synchronize

jukzi:


mpalat (approved):
+1.
Looks good.
However, Could you please add a "this." so before fThread its more explicit/readable since you are already at this? Thanks!

jukzi:
@mpalat i added "this." to all because it looks dump to have that pattern used only for a single member. ok?

SarikaSinha:
@jukzi 
Can you add a description - where are you adding the missing synchronize?
I see that in the first commit you added synchronized on getThread method
https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/259/commits/7020538d0157112f2b368153d351763188441ac1
But I don't see that change in the final commit? I see only this changes.

SarikaSinha (changes requested):
I see only "this" changes.

jukzi:
@SarikaSinha therefore i added two distinct commits in this PR.
in the first commit you see the sychronize and the second purely adding "this.".
you can click on the commits to see there content:
image
image

SarikaSinha (approved):

PR #264, opened 2023-06-14
Avoid reusing of not thread-safe SimpleDateFormat

jukzi:
SimpleDateFormat.format(new Date());
returns same string as immutable and thread-safe
  DateTimeFormatter.format(new Date().toInstant())


mpalat (approved):
+1
Please submit after all the tests are green
Thanks @jukzi

jukzi:
@mpalat there is only an unrelated random fail

org.eclipse.jdt.debug.tests.ui.DebugHoverTests.testBug572629_ChainFieldHover_ArrayLengthChainsOnVariableThisExpression_ExpectValueFromChain

i dont't know how to get that green other then a rebuild (which i can not trigger)and hope for randomness disappear

iloveeclipse:
I've triggered rebuild, will see.

jukzi:
@iloveeclipse thanks. it's gone

Issue #275, opened 2023-07-02
Evaluations inside binary method fails when wild card generics are used 


gayanper:
Take the following sample project, import it into eclipse and replace the following line

`<systemPath><complete_path>/debug-lib/target/debug-lib-1.0-SNAPSHOT.jar</systemPath>` in the debug-app project to a fully qualified path.

Add a break point in the `Observation` line 16 in the debug-lib project. Then debug the Main class in the debug-app. Once the debug point is hit, perform the following evaluation in the debug shell

`new Observation((Object) subject, (Consumer) action)`

**Actual**

The evaluation fails with a compilation error

**Expectation**

The evaluation should return a new instance of the class.

gayanper:
Similar to bug [Bug567801](https://bugs.eclipse.org/bugs/show_bug.cgi?id=567801)

Issue #186, opened 2023-02-20
Auto import the unresolved classes during evaluation


testforstephen:
**Current**:
When the evaluation snippet contains some types not imported in current Java file, it throws compilation errors "unresolved type" during evaluation. To get it work, you have to use the fully qualified name for the missing classes, it is not that convenient on UX perspective.
![image](https://user-images.githubusercontent.com/14052197/220034752-5c8b7882-74cc-43dc-bbf0-e63b7f2e65f8.png)

**Expected**:
Auto import the missing classes when compiling the evaluation snippet.

PR #280, opened 2023-07-07
Fix lack of type resolution issue due to lack of imports. #275

gayanper:
## What it does
The changes will add import statements for found types while traversing symbols in the binary class definition so that the evaluation snippet will properly compiled without needing to have FQN names.

## How to test
Unit tests and sample project

## Author checklist


gayanper:
@testforstephen will you have time to have a look at this ?

testforstephen:
Hi @gayanper,

The approach you use to dynamically add the import statements for the evaluation snippets looks good. I think we can apply it to the case in the issue #186 as well, where we can resolve the unresolved types and automatically append their fully qualified names in the import statements.

However, I have some questions about the use cases you used in the Java8Tests.java file. Both of the snippets `new Observation((Object) subject, (Consumer) action)` and `((Subject)this.subject).getName()` work fine for me with or without this PR.

Could you provide more use cases that demonstrate the problem you want to fix in this PR? Thank you.

testforstephen:
>  Both of the snippets new Observation((Object) subject, (Consumer) action) and ((Subject)this.subject).getName() work fine for me with or without this PR.

It works because I attached `debug-lib-sources.jar` to `debug-lib.jar`, where the eval will use SourceBasedSourceGenerator and works fine. Removing the source attachment can reproduce the issue you mentioned.

However, compared to SourceBasedSourceGenerator, the BinaryBasedSourceGenerator still has some limitation. For example, `new Observation((Object) subject, (Consumer) action)` snippet works fine in the Line 19 breakpoint of Observation for source attached sample, but still not work for binary sample.

gayanper:
> 
> However, compared to SourceBasedSourceGenerator, the BinaryBasedSourceGenerator still has some limitation. For example, `new Observation((Object) subject, (Consumer) action)` snippet works fine in the Line 19 breakpoint of Observation for source attached sample, but still not work for binary sample. 

You mean the snippet already in the unit test or something different?

testforstephen:
I'm using the new unit test added your PR.

gayanper:
@testforstephen do you mean the unit test fail when you execute or it fails when you try run the changes in workspace and try it on a real project debugging scenario?

testforstephen:
> @testforstephen do you mean the unit test fail when you execute or it fails when you try run the changes in workspace and try it on a real project debugging scenario?

The existing test cases you added can pass. My point is that there are other cases not covered well by this PR.

Below is the eval comparison between SourceBasedSourceGenerator and BinaryBasedSourceGenerator.
- The eval snippet `new Observation((Object) subject, (Consumer) action)` works fine if the library source jar exists. SourceBasedSourceGenerator is used when the library source is available.
image

- The eval snippet `new Observation((Object) subject, (Consumer) action)` fails if the library source is unavailable. BinaryBasedSourceGenerator is used if the source is unavailable.
image

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/eval/Java8Tests.java
```diff
@@ -241,4 +241,38 @@ public void testContextEvaluations() throws Exception {
...
 	}
 
+	public void testEvaluate_GH275_StaticBinaryMethod_EvaluateSnippetWithImportedTypes() throws Exception {
+		IJavaThread javaThread = null;
+		try {
+			IJavaLineBreakpoint bp = createLineBreakpoint(getType("com.debug.test.Observation"), 25);
+			javaThread = launchToLineBreakpoint("GH275", bp);
+			assertNotNull("The program did not suspend", javaThread);
+
+			String snippet = "new Observation((Object) subject, (Consumer) action)";
```
gayanper, line 251: @testforstephen isn't this the same snippet you are trying in Eclipse UI which fails for you ?

testforstephen: The snippet is same, but the evaluation context is not same. In this test case, it's stopped at Line 25. The failure case I mentioned is stopped at Line 19. It's supposed to pass as well.

testforstephen (approved):
LGTM

PR #282, opened 2023-07-11
Add ExecutionEnvironmentDescription(Map) constructor and simplify EEDescription implementation

HannesWell:
This PR consists of two commits to enhance `ExecutionEnvironmentDescription`.
1. commits simplifies the code to read EE-files during construction of `ExecutionEnvironmentDescription` and applies a few more minor code clean-ups
2. commit adds a new constructor `ExecutionEnvironmentDescription(Map)` that allows to create a EEDescription from a Map directly.
One use-case for this are for example PDE's API-tools that generate the EE file in (at least) one case, just to let the `EEDescription` read it again.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/environments/ExecutionEnvironmentDescription.java
+++ b/org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/environments/ExecutionEnvironmentDescription.java
@@ -310,7 +310,7 @@ public final class ExecutionEnvironmentDescription {
 				arguments.append(key);
 				if (!value.isEmpty()) {
 					arguments.append('=');
-					value = resolveHome(value, eeHome);
+					value = resolveHome(value, eeHome); // TODO: remove once getProperties() has been removed
 					if (value.indexOf(' ') > -1) {
 						arguments.append('"').append(value).append('"');
 					} else {
@@ -597,7 +597,7 @@ public final class ExecutionEnvironmentDescription {
 	 * @return javadoc location or <code>null</code> if none
 	 */
 	private URL getJavadocLocation() {
-		return EEVMType.getJavadocLocation(fProperties);
+		return EEVMType.getJavadocLocation(this);
 	}
 
 	/**
@@ -607,6 +607,6 @@ public final class ExecutionEnvironmentDescription {
 	 * @since 3.7.0
 	 */
 	private URL getIndexLocation() {
-		return EEVMType.getIndexLocation(fProperties);
+		return EEVMType.getIndexLocation(this);
 	}
 }
--- a/org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/environments/ExecutionEnvironmentDescription.java
+++ b/org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/environments/ExecutionEnvironmentDescription.java
@@ -195,7 +195,19 @@ public final class ExecutionEnvironmentDescription {
 	 * @throws CoreException if unable to read or parse the file
 	 */
 	public ExecutionEnvironmentDescription(File eeFile) throws CoreException {
-		this.fProperties = loadProperties(eeFile);
+		this.fProperties = resolveEEHome(loadProperties(eeFile));
+	}
+
+	/**
+	 * Creates an execution environment description based on the given execution environment description properties. The contained properties are
+	 * defined in <code>https://wiki.eclipse.org/Execution_Environment_Descriptions</code>.
+	 *
+	 * @param eeProperties execution environment description properties
+	 * @throws CoreException if unable to read or parse the file
+	 * @since 3.21
+	 */
+	public ExecutionEnvironmentDescription(Map<String, String> eeProperties) {
+		this.fProperties = resolveEEHome(new LinkedHashMap<>(eeProperties)); // defensive copy
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


HannesWell:
@iloveeclipse or @jukzi, could you please review this?

jukzi:
there are junit errors. probably it's easier to verify if you split it into 2 commits: a) new constructor API b) refactoring

HannesWell:
> there are junit errors. probably it's easier to verify if you split it into 2 commits: a) new constructor API b) refactoring

Thanks for looking at this. It was actually a build error because I forgot to bump the version in the pom.xml as well.
This commit already consist of commits as you suggested, just in reverse order. Performing the refactoring first, makes the subsequent constructor addition simpler.

HannesWell:
@jukzi can we continue on this one?

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/environments/ExecutionEnvironmentDescription.java
```diff
@@ -187,18 +184,35 @@ public final class ExecutionEnvironmentDescription {
...
+	 * @param eeProperties execution environment description properties
+	 * @throws CoreException if unable to read or parse the file
+	 * @since 3.21
+	 */
+	public ExecutionEnvironmentDescription(Map<String, String> eeProperties) {
+		this(new LinkedHashMap<>(eeProperties), null); // defensive copy
+	}
+
+	private ExecutionEnvironmentDescription(Map<String, String> eeProperties, @SuppressWarnings("unused") Object internal) {
+		fProperties = eeProperties; // has to be set before resolveHome() is called
```
jukzi, line 213: better pass eeProperties,.get(EE_HOME) to resolveHome and create a new Map for fProperties. Then the code does not depend on comments

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/environments/ExecutionEnvironmentDescription.java
```diff
@@ -301,26 +308,22 @@ public LibraryLocation[] getLibraryLocations() {
...
-			Entry<String, String> entry = entries.next();
-			String key = entry.getKey();
-			String value = entry.getValue();
+		fProperties.forEach((key, value) -> {
 			boolean appendArgument = !key.startsWith(EE_ARG_FILTER);
 			if (appendArgument) {
 				arguments.append(key);
 				if (!value.isEmpty()) {
 					arguments.append('=');
 					value = resolveHome(value);
```
jukzi, line 317: why is resolveHome called again?

jukzi, line 317: why is resolveHome called again?

jukzi (commented):
i don't know that code enough to make substantial comments.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/environments/ExecutionEnvironmentDescription.java
HannesWell: Done.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/environments/ExecutionEnvironmentDescription.java
HannesWell: Good question.
Propably because one gets the original and modifiable fProperties Map via the `getProperties()` method and can therefore alter the map's content.
I have added another commit to replace all the usages of that method and depreacte `getProperties()` for removal.
That change also replaced all usages of that method in the JDT code-base and AFAICT there is no other use in the Eclipse Platform.

I assume JDT follows the usual two (or was it three years?) deprecation policy before it can be removed? Or do you think that method should be kept?

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/environments/ExecutionEnvironmentDescription.java
```diff
@@ -208,8 +217,11 @@ public ExecutionEnvironmentDescription(File eeFile) throws CoreException {
 	 * have already been performed when resolving property values.
 	 *
 	 * @return properties as a map of {@link String} keys and values
+	 * @deprecated use {@link #getProperty(String)} instead to obtain the value of a key.
 	 */
+	@Deprecated(forRemoval = true)
```
jukzi, line 222: why should this method be deprecated/removed? whats wrong with it?

HannesWell: Since it returns the bare map, it allows to break the property encapsulation and data integrity as described before in https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/282#discussion_r1327963578.

But it is just a suggestion, if you think it should stay, I'm fine with that as well.

jukzi: The way of deprecating in jdt is not easy. I suggest to avoid it.

HannesWell: OK.
Reverted the deprecation but kept the replacement of the few usages so that this method is now unused in the Eclipse Platform.

jukzi (approved):

HannesWell:
Thanks @jukzi.
Does anybody else want to review this or can we merge it?

PR #291, opened 2023-08-03
Fixes #290 - complete the removal of com.ibm.icu dependency

### LINKED ISSUE #290: Remove Dependency on com.ibm.icu
Complete the work of https://bugs.eclipse.org/bugs/show_bug.cgi?id=561910

robstryker:
## What it does

Completes the removal of dependency on com.ibm.icu

## How to test

I am not sure how to test this, however, reading the code for java.text and com.ibm.icu.text shows that getLevelAt should be a reasonable replacement for getLevels(). 

## Author checklist

- [ ] I have thoroughly tested my changes - No, I am not 100% sure how to test this

### DIFF
--- a/org.eclipse.jdt.debug.ui/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.debug.ui/META-INF/MANIFEST.MF
@@ -48,7 +48,6 @@ Require-Bundle: org.eclipse.ui.ide;bundle-version="[3.5.0,4.0.0)",
  org.eclipse.ui.console;bundle-version="[3.4.0,4.0.0)",
  org.eclipse.jdt.core.manipulation;bundle-version="[1.16.0,2.0.0)",
  org.eclipse.search;bundle-version="[3.5.0,4.0.0)",
- com.ibm.icu,
  org.eclipse.ui.forms;bundle-version="[3.4.0,4.0.0)",
  org.eclipse.core.resources,
  org.eclipse.debug.core;bundle-version="[3.12.0,4.0.0)"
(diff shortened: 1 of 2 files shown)


rgrunber:
CC'ing @akurtakov as the comment on https://bugs.eclipse.org/bugs/show_bug.cgi?id=561910#c9 seemed to imply maybe this wasn't possible (did `getLevelAt` not exist then?). Looking at it the change seems very reasonable.

akurtakov:
I think I didn't manage to find it out :) . Back then so many changes went in and in so many places that I have probably overlooked this and probably a good number of other similar cases. I'm pretty sure that a second look will manage to complete far more from the conversion.

robstryker:
I've triggered the bidi code path by modifying java code templates (in window -> preferences)  to display the JavaSourceViewer, and then pasting hebrew or arabic text inside. 

I haven't seen any exceptions yet, so I'm going to assume this works as expected.

rgrunber (commented):
Overall changes looks good. Just a few small things to address. I tried out a mixture of left-to-right and right-to-left, and the L/R arrows keys as well as delete/backspace worked the same way in both.

I wasn't able to reproduce from the Java -> Code Style -> Code Templates editor. However I was able to reproduce with the Java breakpoint editor and confirmed pasting that arrow keys/delete/backspace respected the direction of the character set being used.

@robstryker can you just confirm how you tested with the code templates, because I didn't see how that code path would trigger the JDISourceViewer.

robstryker:
> @robstryker can you just confirm how you tested with the code templates, because I didn't see how that code path would trigger the JDISourceViewer.

I'm sorry if was less than accurate. What I meant is that I tested the same logic that is in this patch by way of another patch on jdt.ui. In jdt.ui, I tested via the JavaSourceViewer viewer using the method described. 

I did not know how to trigger the JDISourceViewer in this plugin so I tried to at least verify that the logic itself was sound by testing it in the other repository.

rgrunber (approved):
Change looks fine to me. The jdt.debug.ui bundle has only had a micro version bump since the 4.28 release (3.13.0 -> 3.13.100) but that should be enough here.

PR #293, opened 2023-08-10
DetectVMInstallationsJob: defer detection until needed.

jukzi:
So that it does not run in jdt.core tests at all

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1190

### DIFF
--- a/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
+++ b/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
@@ -29,6 +29,9 @@ import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.core.runtime.jobs.Job;
+import org.eclipse.core.runtime.preferences.DefaultScope;
+import org.eclipse.core.runtime.preferences.IEclipsePreferences;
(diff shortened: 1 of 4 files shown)


jukzi:
for me it fixes locally running org.eclipse.jdt.core.tests.model.JavaSearchBugsTests.testBug178847()

jukzi:
@mickaelistria please review if that still fulfills the original intend of DetectVMInstallationsJob

jjohnstn:
@jukzi FYI @mickaelistria is on vacation until after Tuesday.

jukzi:
@jjohnstn can you review? You did participate in the initial commit.

jjohnstn:
@jukzi I'll give it a look.  Any idea about the test failures?  Do you need to rebase?

jukzi:
> Any idea about the test failures?

those are only random. i just rebased and had no errors before. See https://ci.eclipse.org/jdt/job/eclipse.jdt.debug-github/job/PR-293/1/

jjohnstn:
Hi @jukzi I tried running the tests locally and they run fine.  I ran into a problem until I switched from using the JustJ JRE to a Java 17 JDK I had.  Is is possible that the JDI tests are finding a bogus VM that has been created by other tests and then it is being discovered?  I am fine with the patch and will approve, but you might want to try disabling the detection when running the JDI tests to see if that fixes the problem.  I have to leave and won't be back in time for the deadline.

jjohnstn (approved):
Looks good.

jukzi:
> I switched from using the JustJ JRE to a Java 17 JDK I had

That sounds like a local error only - which we could fix in future.

Issue #175, opened 2023-01-19
Provide means to query packages exported by the JRE


HannesWell:
In PDE we have a utility method to obtain the packages packages exported by the JRE of a given `IVMInstall/IExecutionEnvironment`, which is duplicated at multiple locations and also has to access JDT-internals:

https://github.com/eclipse-pde/eclipse.pde/blob/47ef407df3082de87f9e8adc26c517204433ce11/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/TargetPlatformHelper.java#L407-L454

I wonder if we should add a corresponding method to `IVMInstall` or `IExecutionEnvironment`, depending on where you think it fits better that returns a Set of package names exported by the JRE.

HannesWell:
> Looking on your snippet, I think `IExecutionEnvironment` is more suitable, at least slightly adopted snippet compiles inside that interface.

In order to be able to query a specific VMInstall for the packages it provides it would be better to add such a method to the `VMInstall` interface. But then API-tools complain that the interface is intended to be implemented by clients and that even a default method is a breaking-change which requires version 4.
If suppressing that error isn't an option a static method could be added to for example `JavaRuntime` that accepts a `VMInstall` and a release and returns the provided packages.

iloveeclipse:
- `testJavaRuntimeQuerySystemPackages_modularJDK` new test fails **differently** on different OS / Java versions.
- `testJavaRuntimeQuerySystemPackages_nonModularJDK` fails because unable to find Java 8 JVM - I would "assume" Java 1.8 as precondition for this test and/or simply only run in Jenkins (where we have the env.variable specified).

@HannesWell: could you please work on that?

Mac Java 17:
https://download.eclipse.org/eclipse/downloads/drops4/I20240424-1800/testresults/html/org.eclipse.jdt.debug.tests_ep432I-unit-macM1-java17_macosx.cocoa.aarch64_17.html

```
begin 0, end -1, length 2

java.lang.StringIndexOutOfBoundsException: begin 0, end -1, length 2
at java.base/java.lang.String.checkBoundsBeginEnd(String.java:4601)
at java.base/java.lang.String.substring(String.java:2704)
at org.eclipse.jdt.debug.tests.core.VMInstallTests.testJavaRuntimeQuerySystemPackages_modularJDK(VMInstallTests.java:249)
```

```
No non-modular VM (<=Java-1.8) found.

[...]

PR #294, opened 2023-08-10
Add API to obtain system-packages provided by a VMInstall

HannesWell:
Fixes https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/175

## What it does
As suggested in https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/175, this adds methods to query the packages provided from a VMInstall.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/VMInstallTests.java
+++ b/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/VMInstallTests.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2013 IBM Corporation and others.
+ * Copyright (c) 2000, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -13,22 +13,36 @@
  *******************************************************************************/
 package org.eclipse.jdt.debug.tests.core;
 
+import static org.junit.Assert.assertThrows;
+
 import java.io.File;
+import java.io.IOException;
 import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.LinkedHashSet;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.NullProgressMonitor;
-import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.Platform;
+import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.debug.testplugin.JavaTestPlugin;
 import org.eclipse.jdt.debug.tests.AbstractDebugTest;
 import org.eclipse.jdt.internal.launching.LaunchingPlugin;
+import org.eclipse.jdt.internal.launching.StandardVMType;
 import org.eclipse.jdt.launching.ILibraryLocationResolver;
 import org.eclipse.jdt.launching.IVMInstall;
 import org.eclipse.jdt.launching.IVMInstall2;
 import org.eclipse.jdt.launching.IVMInstall3;
+import org.eclipse.jdt.launching.IVMInstallType;
 import org.eclipse.jdt.launching.JavaRuntime;
 import org.eclipse.jdt.launching.LibraryLocation;
 import org.eclipse.jdt.launching.VMStandin;
@@ -201,6 +215,128 @@ public class VMInstallTests extends AbstractDebugTest {
 		}
 	}
 
+	private static final Set<String> COMMON_JAVA_PACKAGES = Set.of("java.lang", "java.lang.reflect", "java.util", "java.io");
+	private static final String JAVA9_OR_LATER_PACKAGE = "java.lang.module";
+	private static final String JAVA11_OR_LATER_PACKAGE = "java.net.http";
+	private static final String NON_MODULAR_JDK_PACKAGE = "javax.annotation";
+
+	public void testJavaRuntimeQuerySystemPackages_modularJDK() throws CoreException {
+		IVMInstall vm = JavaRuntime.getDefaultVMInstall();
+		String javaVersion = ((IVMInstall2) vm).getJavaVersion();
+		assertTrue("At least a JDK 11 expected", JavaRuntime.compareJavaVersions(vm, JavaCore.VERSION_11) > 0);
+
+		Set<String> defaultVersionPackages = JavaRuntime.getProvidedVMPackages(vm, null);
+
+		assertContainsAllCommonPackages(vm, defaultVersionPackages);
+		assertTrue(defaultVersionPackages.contains(JAVA9_OR_LATER_PACKAGE));
+		assertTrue(defaultVersionPackages.contains(JAVA11_OR_LATER_PACKAGE));
+		assertFalse(defaultVersionPackages.contains(NON_MODULAR_JDK_PACKAGE));
+
+		Set<String> java9Packages = JavaRuntime.getProvidedVMPackages(vm, "9");
+
+		assertContainsAllCommonPackages(vm, java9Packages);
+		assertTrue(defaultVersionPackages.contains(JAVA9_OR_LATER_PACKAGE));
+		assertFalse(java9Packages.contains(JAVA11_OR_LATER_PACKAGE));
+		assertFalse(java9Packages.contains(NON_MODULAR_JDK_PACKAGE));
+
+		// Test that a null release is equals to the actual version of the VM
+		assertEquals(defaultVersionPackages, JavaRuntime.getProvidedVMPackages(vm, javaVersion));
+
+		// Test failure if non-modular VM is asked for packages of Java<=1.8 or not yet available versions
+		CoreException e1 = assertThrows(CoreException.class, () -> JavaRuntime.getProvidedVMPackages(vm, "1.8"));
+		assertEquals("Cannot query a modular VM (JavaSE-9 or higher) for packages of release: 1.8", e1.getMessage());
+
+		String nextJavaVersion = Integer.toString(Integer.parseInt(javaVersion.substring(0, javaVersion.indexOf('.'))) + 1);
+		CoreException e2 = assertThrows(CoreException.class, () -> JavaRuntime.getProvidedVMPackages(vm, nextJavaVersion));
+		assertEquals("release " + nextJavaVersion + " is not found in the system", e2.getMessage());
+
+		String versionAfterLatestSupported = String.valueOf(Integer.parseInt(JavaCore.latestSupportedJavaVersion()) + 1);
+		CoreException e3 = assertThrows(CoreException.class, () -> JavaRuntime.getProvidedVMPackages(vm, versionAfterLatestSupported));
+		// Passing a release not yet supported by JDT should not fail if the JDK actually provides it (e.g. if one uses early-access builds).
+		// Since EA-builds are not generally available in all test setups we check at least that the method passes the initial validation
+		assertEquals("release " + versionAfterLatestSupported + " is not found in the system", e3.getMessage());
+
+		CoreException e4 = assertThrows(CoreException.class, () -> JavaRuntime.getProvidedVMPackages(vm, "definitivly-not-a-version"));
+		assertEquals("Invalid release: definitivly-not-a-version", e4.getMessage());
+	}
+
+	public void testJavaRuntimeQuerySystemPackages_nonModularJDK() throws Exception {
+		try (AutoCloseableSupplier<IVMInstall> nonModularVM = searchFirstNonModularVM()) {
+			IVMInstall vm = nonModularVM.get();
+
+			Set<String> packages = JavaRuntime.getProvidedVMPackages(vm, null);
+
+			assertContainsAllCommonPackages(vm, packages);
+			assertTrue("Not found: " + NON_MODULAR_JDK_PACKAGE, packages.contains(NON_MODULAR_JDK_PACKAGE));
+			assertFalse("Found unexpected package: " + JAVA9_OR_LATER_PACKAGE, packages.contains(JAVA9_OR_LATER_PACKAGE));
+			assertFalse("Found unexpected package: " + JAVA11_OR_LATER_PACKAGE, packages.contains(JAVA11_OR_LATER_PACKAGE));
+
+			// Test that for non-modular VMs the release information is just ignored
+			assertEquals(packages, JavaRuntime.getProvidedVMPackages(vm, "1.6"));
+			assertEquals(packages, JavaRuntime.getProvidedVMPackages(vm, "definitivly-not-a-version"));
+		}
+	}
+
+	private void assertContainsAllCommonPackages(IVMInstall vm, Set<String> packages) {
+		Set<String> missing = new LinkedHashSet<>(COMMON_JAVA_PACKAGES);
+		missing.removeAll(packages);
+		assertTrue("Not all packages found in " + vm.getInstallLocation() + ", missing " + missing, missing.isEmpty());
+	}
+
+	private AutoCloseableSupplier<IVMInstall> searchFirstNonModularVM() throws IOException {
+		IVMInstallType vmType = JavaRuntime.getVMInstallType(StandardVMType.ID_STANDARD_VM_TYPE);
+
+		Set<File> candidateDirectories;
+		String nonModularJavaHome = System.getenv("NON_MODULAR_JAVA_HOME");
+		if (nonModularJavaHome != null) {
+			System.out.println(getName() + ": NON_MODULAR_JAVA_HOME: " + nonModularJavaHome);
+			candidateDirectories = Set.of(new File(nonModularJavaHome));
+		} else {
+			Path defaultVMLocation = JavaRuntime.getDefaultVMInstall().getInstallLocation().toPath();
+			int maxSearchDepth = 1;
+			Path root = Stream.iterate(defaultVMLocation, Objects::nonNull, Path::getParent) //
+					.skip(maxSearchDepth).findFirst().orElseThrow();
+			try (Stream<Path> directories = Files.walk(root, maxSearchDepth);) {
+				candidateDirectories = directories.map(Path::toFile).collect(Collectors.toSet());
+			}
+			Arrays.stream(JavaRuntime.getVMInstallTypes()).flatMap(t -> Arrays.stream(t.getVMInstalls())) //
+					.map(IVMInstall::getInstallLocation) //
+					.forEach(candidateDirectories::remove);
+		}
+		System.out.println(getName() + ": found JVM candidates: " + candidateDirectories);
+
+		IVMInstall candidate = null;
+		for (File file : candidateDirectories) {
+			if (vmType.validateInstallLocation(file).isOK()) {
+				String id = "test-vm-" + System.nanoTime();
+				IVMInstall vmCandidate = vmType.createVMInstall(id);
+				vmCandidate.setInstallLocation(file);
+				if (!JavaRuntime.isModularJava(vmCandidate)) {
+					candidate = (vmCandidate);
+					break;
+				}
+				vmType.disposeVMInstall(id);
+			}
+		}
+		assertNotNull("No non-modular VM (<=Java-1.8) found.", candidate);
+		IVMInstall vm = candidate;
+		System.out.println(getName() + ": found non-modular JVM: " + vm.getInstallLocation());
+		return new AutoCloseableSupplier<>() {
+			@Override
+			public IVMInstall get() {
+				return vm;
+			}
+
+			@Override
+			public void close() throws Exception {
+				vmType.disposeVMInstall(vm.getId());
+			}
+		};
+	}
+
+	private interface AutoCloseableSupplier<T> extends AutoCloseable, Supplier<T> {
+	}
+
 	/**
 	 * Checks the given {@link LibraryLocation}s to ensure they reference the testing resolver paths
 	 */
@@ -231,7 +367,7 @@ public class VMInstallTests extends AbstractDebugTest {
 	 * @return the {@link VMStandin}
 	 */
 	VMStandin getEEStandin(String filename) throws CoreException {
-		File ee = JavaTestPlugin.getDefault().getFileInPlugin(new Path(filename));
+		File ee = JavaTestPlugin.getDefault().getFileInPlugin(IPath.fromOSString(filename));
 		assertNotNull("The EE file "+filename+" was not found", ee);
 		VMStandin vm = JavaRuntime.createVMFromDefinitionFile(ee, "resolver-ee", "resolver-ee-id");
 		assertNotNull("the VM standin should exist for "+filename, vm);
--- a/org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/AbstractVMInstall.java
+++ b/org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/AbstractVMInstall.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2023 IBM Corporation and others.
+ * Copyright (c) 2000, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -21,6 +21,8 @@ import java.io.IOException;
 import java.net.URL;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 import javax.xml.parsers.DocumentBuilder;
 
@@ -62,6 +64,7 @@ public abstract class AbstractVMInstall implements IVMInstall, IVMInstall2, IVMI
 	private volatile LibraryLocation[] fSystemLibraryDescriptions;
 	private volatile URL fJavadocLocation;
 	private volatile String fVMArgs;
+	final Map<String, Set<String>> systemPackages = new ConcurrentHashMap<>();
 	/**
 	 * Map VM specific attributes that are persisted restored with a VM install.
 	 * @since 3.4
--- a/org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
+++ b/org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
@@ -13,6 +13,7 @@
  *     Frits Jalvingh - Contribution for Bug 459831 - [launching] Support attaching
  *     	external annotations to a JRE container
  *     Ole Osterhagen - Issue 327 - Attribute "Without test code" is ignored in the launcher
+ *     Hannes Wellmann - Add API to obtain system-packages provided by a VMInstall
  *******************************************************************************/
 package org.eclipse.jdt.launching;
 
@@ -34,12 +35,16 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Hashtable;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
 
 import javax.xml.parsers.DocumentBuilder;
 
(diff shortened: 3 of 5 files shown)


org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
```diff
@@ -3449,6 +3452,97 @@ public static VMStandin createVMFromDefinitionFile(File eeFile, String name, Str
...
+					try (ZipFile zip = new ZipFile(path.toFile())) {
+						return zip.stream().filter(e -> !e.isDirectory()).map(ZipEntry::getName).filter(n -> !n.startsWith("META-INF/")).map(n -> { //$NON-NLS-1$
+							int i = n.lastIndexOf('/');
+							return n.substring(0, i);
+						}).distinct().map(n -> n.replace('/', '.')).sorted().toList();
+					} catch (IOException e) {
+						throw new CoreException(Status.error("Failed to read packages in JVM library for " + vm + ", at " + path, e)); //$NON-NLS-1$//$NON-NLS-2$
+					}
+				}
+			}
```
HannesWell, line 3530: For non-modular, pre Java-8 VMs I assume JDT already has some code to read all the VM-jars (like rt.jar, resources.jar, ...), but I didn't manage to find that until now.
I'll continue searching, but if anybody has a hint, that's more than welcome.

HannesWell:
@iloveeclipse, @jukzi, could you please review this?

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/AbstractVMInstall.java
```diff
@@ -43,6 +44,7 @@
 import org.eclipse.jdt.internal.launching.Standard11xVM;
 import org.eclipse.jdt.internal.launching.StandardVMType;
 import org.eclipse.osgi.util.NLS;
+import org.graalvm.nativeimage.Platform;
```
jukzi, line 47: import gone wrong

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
```diff
@@ -3449,6 +3452,97 @@ public static VMStandin createVMFromDefinitionFile(File eeFile, String name, Str
...
+	 * @param vm
+	 *            the vm-install to query
+	 * @param release
+	 *            the release to restrict the packages for, ignored for non-modular VMs
+	 * @return the distinct and alphabetically sorted collection of names of all packages in the VM
+	 * @throws CoreException
+	 *             if an error occurs or a modular VM is query for packages of Java-8 or lower
+	 * @since 3.21
+	 */
+	@SuppressWarnings("restriction")
```
jukzi, line 3494: use RuntimeException.getCause instead

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
```diff
@@ -3449,6 +3452,97 @@ public static VMStandin createVMFromDefinitionFile(File eeFile, String name, Str
...
+	 * If the given VM-Install is {@link #isModularJava(IVMInstall) non modular}, the execution-environment is ignored.
+	 * </p>
+	 *
+	 * @param vm
+	 *            the vm-install to query
+	 * @param environment
+	 *            the EE to restrict the packages for, ignored for non-modular VMs
+	 * @return the distinct and alphabetically sorted collection of names of all packages in the VM
+	 * @throws CoreException
+	 *             if an error occurs or a modular VM is query for packages of Java-8 or lower
```
jukzi, line 3469: "is query" sounds wrong.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
```diff
@@ -3449,6 +3452,97 @@ public static VMStandin createVMFromDefinitionFile(File eeFile, String name, Str
...
+						throw new CoreException(Status.error("Failed to read packages in JVM library for " + vm + ", at " + path, e)); //$NON-NLS-1$//$NON-NLS-2$
+					}
+				}
+			}
+		}
+		if (JavaCore.compareJavaVersions(release, JavaCore.VERSION_1_8) <= 0) {
+			throw new UnsupportedOperationException("Cannot query a modular VM (JavaSE-9 or higher) for packages of JavaSE-1.8 and lower"); //$NON-NLS-1$
+		}
+		String path = new File(vm.getInstallLocation(), JRT_PATH).toString();
+		var jrt = org.eclipse.jdt.internal.core.builder.ClasspathLocation.forJrtSystem(path, null, null, release);
```
jukzi, line 3536: i don't like the use of jdt internals in debug

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/AbstractVMInstall.java
HannesWell: Ups. Probably went wrong during rebase. Fixed it. Thanks.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
HannesWell: That's right. Fixed it.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
HannesWell: Ok. Done.
In general I tried to be as specific as possible so that really only the `CoreException` thrown within the `computeIfAbsent` is unwrapped.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
HannesWell: Me neither, but I don't have a better idea. If you have one, please let me know.
Besides that there are already many uses of jdt internals in this class and I think it is better done here than in PDE or somewhere else even further away from jdt-core.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
HannesWell: @jukzi any idea for this one?

jukzi: maybe something like org.eclipse.jdt.internal.core.builder.ClasspathJar.readJarContent(SimpleSet)

HannesWell: Thanks for that hint.
I looked into it and tried `new ClasspathJar(zip, null, false)listPackages();`, but that returns a list that besides the packages that contain classes also contains all the possible prefixes of those packages with actual content.
For example for `java.util` this also returns just `java`.
But that is not wanted here and I don't see a way to filter them.

Furthermore ClasspathJar internally uses a PackageCache, which I'm a bit hesitate to use in the shade of https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1526, especially since the method implemented here already uses its own cache.

HannesWell:
@jukzi, @iloveeclipse what do you think about the current state of the PR?
The open points are IMHO not ideal, but as written in the corresponding threads, I've not found a better solution.

In the light of https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1526, I wonder if it would be reasonable to query a vm for packages without caching the JRT-FS?
I'm not familiar with them at all, but could it furthermore be only a little overhead (performance and memory wise) if for example for all available the packages are read? Then all releases could be read as batch and then the FS could be closed again. But this is just an idea.

SarikaSinha (dismissed):
Can you add a test case to confirm that it is indeed fetching the right thing?

We can create our own constant rathe than using SUFFIX_STRING_class as it is a very generic constant not likely to change.

jukzi:
@SuppressWarnings("restriction") could be avoided by declaring jdt.debug as x-friend in jdt.core.

jukzi (commented):
For returning a sorted Collection it is more convenient  to use a List, as a Set has no order by default. That the result contains only distinct values is intuitive even if its a List.
I am confused by the title "Provide means". We use to speak about providing "API".

HannesWell:
> For returning a sorted Collection it is more convenient to use a List, as a Set has no order by default. That the result contains only distinct values is intuitive even if its a List. 

A collection with a distinct iteration sequence does not necessarily provide access by an index (which is the key-point of a list), which is why we have a `SequencedSet` in Java-21 (would have used that if we were on 21 already). At the same time a list does not have to be sorted. And while a list can have distinct entries a set can also have a fixed iteration order.
So both could somehow fit (but would require more specification in the doc) and I also contemplated to use List. In the end choose Set because I don't think users of this API are interested about a specific index of a package but more if a VM provides a specific package and then Set.contains() is much faster with O(1). 
Iteration over the returned collection, which is at least use case I have for PDE, has the same runtime for both types.
Therefore I think a Set is more suitable.

> I am confused by the title "Provide means". We use to speak about providing "API".

Indeed. Adjusted the title, it used to be the issue title.

> > In the light of [eclipse-jdt/eclipse.jdt.core#1526](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1526), I wonder if it would be reasonable to query a vm for packages without caching the JRT-FS?
> > I'm not familiar with them at all, but could it furthermore be only a little overhead (performance and memory wise) if for example for all available the packages are read? Then all releases could be read as batch and then the FS could be closed again. But this is just an idea.
> 
> I also looked a little bit into this and it would probably be doable, but would require some cooperation in JDT-core. In order to achieve that we could just add a (internal) method to `ClasspathLocation` or `JRTUtil` that is tailored to this methods need and just returns the packages exported to the public by the given VM for a given release. That method could then avoid caching and could only read what's for this goal.

I have now looked into this in more detail and most of the implementation actually re-uses the same 'large' objects so I don't think there is really a lot to save. The only little thing I found is https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2370. But that can be discussed independently from this PR.

Jenkinsfile
```diff
@@ -12,10 +12,13 @@ pipeline {
 		maven 'apache-maven-latest'
 		jdk 'openjdk-jdk21-latest'
 	}
+	environment {
+		NON_MODULAR_JAVA_HOME = "${tool(type: 'jdk, name: 'temurin-jdk8-latest')}"
+	}
```
HannesWell, line 17: In order to pick this change up in a verification build (and make all tests pass) a Committer of JDT would have to toe _Replay_ the build with this change.

HannesWell:
Any chance we can land this for M2 (maybe with sufficient time for PDE to pick it up)?

@iloveeclipse do you want to review this as well?

iloveeclipse:
@SarikaSinha : I assume you could re-review the PR & change (or discard) your vote because a test was added as requested.

@HannesWell : I'm looking into it. I will push changes which I believe necessary directly to the PR. One thing I see immediately is API tools error because of too old bundle version used in "since" tag.

iloveeclipse:
Note: while reviewing I've found a regression that breaks the test, so I've rebased this PR on top of https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/437 which I plan to merge ASAP.

Beside this, I've fixed issues I've found so far & pushed changes here.

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/VMInstallTests.java
```diff
@@ -201,6 +217,114 @@ public void testLibraryResolver4() throws Exception {
...
+
+	public void testJavaRuntimeQuerySystemPackages_nonModularJDK() throws Exception {
+		try (AutoCloseableSupplier<IVMInstall> nonModularVM = searchFirstNonModularVM()) {
+			IVMInstall vm = nonModularVM.get();
+
+			Set<String> packages = JavaRuntime.getProvidedVMPackages(vm, null);
+
+			LinkedHashSet<String> missing = new LinkedHashSet<>(COMMON_JAVA_PACKAGES);
+			missing.removeAll(packages);
+			assertTrue("Not all packages found in " + vm.getInstallLocation() + ", missing " + missing, missing.isEmpty());
```
iloveeclipse, line 262: Changed assertions to provide meaningful fail message if some package is missing or not present

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/VMInstallTests.java
```diff
@@ -201,6 +217,114 @@ public void testLibraryResolver4() throws Exception {
...
+				if (!JavaRuntime.isModularJava(vmCandidate)) {
+					candidateVm = Optional.of(vmCandidate);
+					break;
+				}
+				vmType.disposeVMInstall(id);
+			}
+		}
+		assertTrue("No non-modular VM (<=Java-1.8) found.", candidateVm.isPresent());
+		IVMInstall vm = candidateVm.get();
+		System.out.println(getName() + ": found non-modular JVM: " + vm.getInstallLocation());
```
iloveeclipse, line 311: I've added few system outs to print out what we've checked and what we finally found.

HannesWell:
> The test below fails because it can't find non modular JVM. I'm pretty sure there should be some installed, so I guess changes on Jenkinsfile either not working (env is not passed) or path is different. The test case could add this info to the fail message.

I'm short in time at the moment, but as a quick note:
This should be the cause of the current failure: https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/294#discussion_r1572744031

Thank you for for your help!

iloveeclipse:
`testJavaRuntimeQuerySystemPackages_nonModularJDK:` found JVM candidates: [/opt/tools/java/openjdk/jdk-21, /opt/tools/java/openjdk/jdk-21/latest]`
Looks like jenkins environment variable is not properly set.

iloveeclipse:
> This should be the cause of the current failure: [#294 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/294#discussion_r1572744031)

I've triggered https://ci.eclipse.org/jdt/job/eclipse.jdt.debug-github/job/PR-294/16/ with the environment change

iloveeclipse:
> > This should be the cause of the current failure: [#294 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/294#discussion_r1572744031)
> 
> I've triggered https://ci.eclipse.org/jdt/job/eclipse.jdt.debug-github/job/PR-294/16/ with the environment change

Test succeeded: https://ci.eclipse.org/jdt/job/eclipse.jdt.debug-github/job/PR-294/16/testReport/junit/org.eclipse.jdt.debug.tests.core/VMInstallTests/testJavaRuntimeQuerySystemPackages_nonModularJDK/

```
!MESSAGE [testJavaRuntimeQuerySystemPackages_nonModularJDK] setUp
testJavaRuntimeQuerySystemPackages_nonModularJDK: NON_MODULAR_JAVA_HOME: /opt/tools/java/temurin/jdk-8/latest
testJavaRuntimeQuerySystemPackages_nonModularJDK: found JVM candidates: [/opt/tools/java/openjdk/jdk-21, /opt/tools/java/temurin/jdk-8/latest, /opt/tools/java/openjdk/jdk-21/latest]
testJavaRuntimeQuerySystemPackages_nonModularJDK: found non-modular JVM: /opt/tools/java/temurin/jdk-8/latest
```

iloveeclipse (approved):
LGTM

iloveeclipse:
@SarikaSinha, all : any objections to merge? 

The `OpenFromClipboardTests`that is failing is known to be a random issue, the new tests are fine: https://ci.eclipse.org/jdt/job/eclipse.jdt.debug-github/job/PR-294/16/testReport/junit/org.eclipse.jdt.debug.tests.core/VMInstallTests/

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/VMInstallTests.java
```diff
@@ -201,6 +216,106 @@ public void testLibraryResolver4() throws Exception {
...
 	}
 
+	private static final Set<String> COMMON_JAVA_PACKAGES = Set.of("java.util", "java.io"); // TODO: TBD
+	private static final String JAVA9_OR_LATER_PACKAGE = "java.lang.module";
+	private static final String JAVA11_OR_LATER_PACKAGE = "java.net.http";
+	private static final String NON_MODULAR_JDK_PACKAGE = "javax.annotation";
+
+	public void testJavaRuntimeQuerySystemPackages_modularJDK() throws CoreException {
+		IVMInstall vm = JavaRuntime.getDefaultVMInstall();
+		String javaVersion = ((IVMInstall2) vm).getJavaVersion();
```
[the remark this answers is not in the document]
HannesWell: Good point. Thanks for fixing that.

> Good that we have the test here :-)

Absolutely and this is not the first issue found by the tests. I already fixed one regarding null keys immediately after creating them.

Jenkinsfile
```diff
@@ -12,10 +12,14 @@ pipeline {
 		maven 'apache-maven-latest'
 		jdk 'openjdk-jdk21-latest'
 	}
+	environment {
+		NON_MODULAR_JAVA_HOME = tool(type:'jdk', name:'temurin-jdk8-latest')
```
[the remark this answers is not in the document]
HannesWell: Indeed the closing `'` was missing. With that when replaying the change the previous variant would probably have worked as well, but this has a few characters less and is easier to read so I'm happy to keep it without enclosing `"${`.

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/VMInstallTests.java
```diff
@@ -201,6 +217,114 @@ public void testLibraryResolver4() throws Exception {
...
+			candidateDirectories.add(new File(nonModularJavaHome));
+			System.out.println(getName() + ": NON_MODULAR_JAVA_HOME: " + nonModularJavaHome);
+		}
+
+		System.out.println(getName() + ": found JVM candidates: " + candidateDirectories);
+
+		Optional<IVMInstall> candidateVm = Optional.empty();
+		for (File file : candidateDirectories) {
+			if (vmType.validateInstallLocation(file).isOK()) {
+				String id = "test-vm-" + System.nanoTime();
```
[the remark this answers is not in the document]
HannesWell: I'm fine with this change, but where did it miss the disposal?

HannesWell:
Thank you @iloveeclipse for all your fixes, that fixed some sharp edges.

I have enhanced the test-code further to remove a resolved TODO, dead code and to use the more meaning full assertions also for the modular case.

I'm fine with everything, but wonder if we should still search for other VMs if `NON_MODULAR_JAVA_HOME` is set? I would say, if that is set only that should be used and the search should be avoided. That way one can even specify one exact JDK to use for the test (not that I now a case where a specific Java<=1.8 JDK is necessary).

iloveeclipse:
> but wonder if we should still search for other VMs if `NON_MODULAR_JAVA_HOME` is set? I would say, if that is set only that should be used and the search should be avoided.

As you like, I have no preference, added that just in case Jenkins would find some 1.x JVMs which seem to be not the case anyway.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
```diff
@@ -3460,6 +3465,104 @@ public static VMStandin createVMFromDefinitionFile(File eeFile, String name, Str
...
+	public static Set<String> getProvidedVMPackages(IVMInstall vm, String release) throws CoreException {
+		boolean isModular = JavaRuntime.isModularJava(vm);
+		if (release != null) {
+			// normalize release (from 17.0.6 to 17)
+			release = org.eclipse.jdt.internal.compiler.impl.CompilerOptions.versionFromJdkLevel(org.eclipse.jdt.internal.compiler.impl.CompilerOptions.versionToJdkLevel(release));
+		}
+		// If release is empty we assume no release info available
+		if (release != null && release.isEmpty()) {
+			release = null;
+		}
```
HannesWell, line 3498: I wonder if we should just throw an exception if the `release` value is not as expected? We could even consider to throw one if a non modular `release`, i.e. `1.8` or less is specified since they are effectively ignored (but that might require too much logic on the call side)?
If I understand the code called correctly it will result in an empty string and eventually `null` if one just specifies any non-sense.

HannesWell:
> > but wonder if we should still search for other VMs if `NON_MODULAR_JAVA_HOME` is set? I would say, if that is set only that should be used and the search should be avoided.
> 
> As you like, I have no preference, added that just in case Jenkins would find some 1.x JVMs which seem to be not the case anyway.

Then lets only search if the variable is not set.

Yes finding any 1.x JVM would require a search depth of 3 due to the way the JDKs are structured on the EF Jenkins and because there is only an openJDK 9 or later:
https://wiki.eclipse.org/Jenkins#JDK

JDT-Debug currently uses `openjdk-jdk21-latest` by default. But only Temurin, Adopt, Oracle or IBM provide JDK <=8.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
iloveeclipse: May I ask you to try with Java 23 release (I haven't tried)? JDT doesn't fully support it yet, so I believe the code above will return empty string, and the question would be: if so, would it make sense to continue (assuming it is the JVM we are running Eclipse) or throw an error? If we continue, would the client of this API get what is expected or not?

HannesWell: If I call for example `JavaRuntime.getProvidedVMPackages(vm, "24")`
Then the following code is called:
https://github.com/eclipse-jdt/eclipse.jdt.core/blob/279fc9d3e2dfbf00a30b2046ee710dfdc84aaeab/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/impl/CompilerOptions.java#L909-L912

Which eventually returns `ClassFileConstants.MAJOR_LATEST_VERSION` that currently corresponds to Java 22.
So that does not fail.
Thinking about it again, I think a suitable compromise would be to throw an exception if `CompilerOptions.versionToJdkLevel()` returns 0. This would permit JVM versions like `17.0.10` which is then mapped to spec verison `17`, but it would reject random other strings as input.

The only problem I see with this approach is that 23 is currently mapped to 22, even if the targeted JVM is e.g. a 23 EA build and could be queried. We should actually just check if the passed release argument is a version and leave the check if that version is supported to the JRT filesystem.

iloveeclipse: > a suitable compromise would be to throw an exception if CompilerOptions.versionToJdkLevel() returns 0.

My concern is that people would not be able to use this API running on latest JVM not yet supported by JDT.
I assume you plan to use it in PDE, so what would break in that case, assuming they "target" JVM will be "known" by JDT, so that in theory we shouldn't fail in this case?

HannesWell: > > a suitable compromise would be to throw an exception if CompilerOptions.versionToJdkLevel() returns 0.
> 
> My concern is that people would not be able to use this API running on latest JVM not yet supported by JDT. I assume you plan to use it in PDE, so what would break in that case, assuming they "target" JVM will be "known" by JDT, so that in theory we shouldn't fail in this case?

Yes my plan is to use that in PDE and in general if querying the System-Packages fails it would either mean that for the corresponding EE no system-packages are available or we just not provide the EE at all (the exact behavior is to be defined). At the moment only the EEs defined by JDT are available in PDE's state anyway.
That being said, for PDE this new API should be called in a way that is always correct and never throws an exception (at this point).
My intention for throwing an exception is for yet not known other users that should get some feed back if they pass unsuitable values.

But with the current approach fetching the system-packages for Java-23 is possible even if one has a JDK 23 installed, because 23 is converted to 22 as described before although the `JavaSE-23` EE already exists: https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/426
Therefore the current approach is not suitable in this case.

I'm currently thinking about a more fine tuned logic.

iloveeclipse:
Hannes, if you extract Jenkinsfile change to dedicated PR I could merge it so this one could be properly tested.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
HannesWell: I have just pushed a second commit to this branch with a suggestion that should prevent the biggest mistakes but still allows to query JDKs of version that are not yet supported by ECJ/java-core.
The second commit also contains corresponding adjustments/extensions of the tests.

If you are fine with them that commit should be squashed into the first.

HannesWell:
> Hannes, if you extract Jenkinsfile change to dedicated PR I could merge it so this one could be properly tested.

Yes. See https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/438.

iloveeclipse:
> > Hannes, if you extract Jenkinsfile change to dedicated PR I could merge it so this one could be properly tested.
> 
> Yes. See #438.

Now please rebase.

HannesWell:
> > > Hannes, if you extract Jenkinsfile change to dedicated PR I could merge it so this one could be properly tested.
> > 
> > 
> > Yes. See #438.
> 
> Now please rebase.

Done.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
```diff
@@ -3460,6 +3465,120 @@ public static VMStandin createVMFromDefinitionFile(File eeFile, String name, Str
...
+					throw coreException; // throw passed through exception
+				}
+				throw e;
+			}
+		}
+		return querySystemPackages(vm, release, isModular);
+	}
+
+	private static String normalizeRelease(String release, boolean isModularVM) throws CoreException {
+		if (release != null) {
```
iloveeclipse, line 3514: Could we please invert condition to avoid that entire method body is inside a single "if"?

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
```diff
@@ -3460,6 +3465,120 @@ public static VMStandin createVMFromDefinitionFile(File eeFile, String name, Str
...
+			@SuppressWarnings("restriction")
+			long releaseJdkLevel = org.eclipse.jdt.internal.compiler.impl.CompilerOptions.versionToJdkLevel(release);
+			if (releaseJdkLevel == 0) {
+				// Ensure the specified release is a version (which may have minor or micro segments).
+				// Note that versionToJdkLevel limits the version to the latest one supported by ECJ, which can be less than the JDK provides.
+				throw new CoreException(Status.error("Invalid release: " + release)); //$NON-NLS-1$
+			}
+			if (!isModularVM) {
```
iloveeclipse, line 3522: should this be moved before the previous check? We don't care about content in this case anyway.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
```diff
@@ -3460,6 +3465,120 @@ public static VMStandin createVMFromDefinitionFile(File eeFile, String name, Str
...
+			if (JavaCore.compareJavaVersions(release, JavaCore.VERSION_1_8) <= 0) {
+				throw new CoreException(Status.error("Cannot query a modular VM (JavaSE-9 or higher) for packages of release: " + release)); //$NON-NLS-1$
+			}
+			// normalize release (e.g. from 17.0.6 to 17)
+			int firstDotIndex = release.indexOf('.');
+			if (firstDotIndex > -1) {
+				release = release.substring(0, firstDotIndex);
+			}
+			if (!release.chars().allMatch(Character::isDigit)) {
+				throw new CoreException(Status.error("Invalid release : " + release)); //$NON-NLS-1$
```
iloveeclipse, line 3534: please remove space before colon in  `release :`

HannesWell:
> Hannes, if you extract Jenkinsfile change to dedicated PR I could merge it so this one could be properly tested.

Now the build passed.
Please note that there is still a commit to squash, just in case you are perfectly fine with it :)

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
iloveeclipse: Yes, I think it is OK, so please squash

HannesWell:
Implemented all your remarks and squashed the second commit into the first.

HannesWell:
Thank you Andrey for your reviews and help.

> I plan to merge this PR for the next IBuild (will be 4.32 M2) tomorrow, assuming there will be no objections before 23:00 CET today.

Since I have no chance to pick this up in PDE for M2 for me there is no urgency to submit it today. Nevertheless I have no objections to have this sooner than later :).

iloveeclipse:
Let merge then.

SarikaSinha:
Missed adding my comment yday, Looks good to me.

Issue #320, opened 2023-10-06
System hangs after 2x resuming from "SUSPEND_VM" breakpoint


raghucssit:
If we have multiple `Suspend VM` breakpoint and all are executed at the end `Lauch Application` goes infinte and need to be terminated manually.
Please use the below snippet to reproduce the issue:
```
public class SuspendVm {
    public static void main(String[] args) {
    	Runtime.getRuntime().addShutdownHook(new Thread("Printing Shutdown Hook") {
    		@Override
    		public void run() {
    			System.out.println("Shutdown called!");
    		}
    	});
        System.out.print(""); // set a "suspend VM" breakpoint here, and resume *thread*
        main();
        lastCall();
    }
	public static void main() {
    	System.out.println(""); // set a "suspend VM" breakpoint here, and resume *VM*
    }
    private static void lastCall() {
		System.out.println("Going to shutdown?"); // set a "default" breakpoint here, and resume *thread*
	}
}
```

Problem here is when each `Suspend VM` is hit it's count on JDB increases. In debugger UI selection remains on `Thread` on resume thread is resumed. But we have to initiate a `Resume` for every `Suspend` for `Virtual Machine` just like we do for `Threads`. This implmentation is missing.

PR #321, opened 2023-10-06
Fix virtual machine hangs after 2x suspend vm breakpoint

raghucssit:
We need to initiate equal number of resume for suspend vm requests by threads.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/320

### DIFF
--- a/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java
+++ b/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java
@@ -239,6 +239,9 @@ public class JDIDebugTarget extends JDIDebugElement implements
 	 */
 	private int fSuspendCount;
 
+	/** Number of extra VM.resume operations needed due SUSPEND_VM breakpoints */
+	private final AtomicInteger extraResumeCount = new AtomicInteger(0);
+
 	/**
 	 * Evaluation engine cache by Java project. Engines are disposed when this
 	 * target terminates.
@@ -1340,6 +1343,12 @@ public class JDIDebugTarget extends JDIDebugElement implements
 			resumeThreads();
 			VirtualMachine vm = getVM();
 			if (vm != null) {
+				// Extra resume call per each VM_SUSPEND breakpoint hit before
+				while (extraResumeCount.getAndUpdate(count -> count > 0 ? --count : 0) > 0) {
+					vm.resume();
+				}
+
+				// "regular" resume
 				vm.resume();
 			}
 			if (fireNotification) {
@@ -1770,6 +1779,9 @@ public class JDIDebugTarget extends JDIDebugElement implements
 		} else {
 			fireSuspendEvent(DebugEvent.BREAKPOINT);
 		}
+		// We need to remember how many times VM was suspended, to call
+		// resume as many times as needed
+		extraResumeCount.incrementAndGet();
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


raghucssit:
Changed author as per my eclipse account.
Rebased with master.

org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java
```diff
@@ -334,6 +334,10 @@ public class JDIDebugTarget extends JDIDebugElement implements
 	 */
 	private final Map<Long, String> objectLabels = new HashMap<>();
 
+	private int suspendCount;
```
iloveeclipse, line 337: Debugger is always multi-threaded, so all variables should be volatile

org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java
```diff
@@ -1340,6 +1344,13 @@ protected void resume(boolean fireNotification) throws DebugException {
 			resumeThreads();
 			VirtualMachine vm = getVM();
 			if (vm != null) {
+				if (pendingResumeRequests) {
+					while (suspendCount > 1) {
```
iloveeclipse, line 1348: `suspendCount`, once set from zero to `> 1`, will stay always `>= 1` after this while loop.
Why not check `while (suspendCount > 0)`?

org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java
```diff
@@ -1340,6 +1344,13 @@ protected void resume(boolean fireNotification) throws DebugException {
...
+						vm.resume();
+						suspendCount--;
+						pendingResumeRequests = false;
+					}
+				}
```
iloveeclipse, line 1353: Shouldn't `vm.resume();` call below be in the `else` block?

iloveeclipse (changes requested):
Hard to play with / verify the patch without test case.

Please add a test case to `org.eclipse.jdt.debug.tests.breakpoints.SuspendVMBreakpointsTests`

org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java
raghucssit: This is a default `vm.resume();` which was existing.
When the `suspendCount  = 1` this works by default.
If there were more than 1 Suspend VM breakpoints are set they will be resumed inside while loop.
Whenever a `IJavaBreakpoint.SUSPEND_VM` breakpoint is hit `JDIDebugTarget#suspendedByBreakpoint(JavaBreakpoint breakpoint, boolean queueEvent, EventSet set)` is invoked.
So I am counting the number of Suspend VM breakpoints hit. And whenever VM(Debug Target) Resume is requested all the pending VM suspend will be cleared inside while loop.

org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIDebugTarget.java
iloveeclipse: I've fixed that in my last pushed commit.

iloveeclipse (approved):

raghucssit:
> @raghucssit : I've reworked PR a bit, but the main idea is same & original fix was in right place, so main idea untouched.
> 
> * get rid of the extra boolean flag
> * renamed counter and changed it to a thread safe AtomicInteger
> * added comments in code
> * slightly changed test
> 
> Please review.

Code changes looks good.
I have pulled and did some manual testing, Everything works as expected.
Good to merge.

Issue #323, opened 2023-10-13
DetectVMInstallationsJob open network connection


nitind:
On Eclipse 2023-09 (4.29), Temurin OpenJDK 17.0.4.1, macOS

In spite of the parser settings applied, the DetectVMInstallationsJob attempts to retrieve the DTD, which means opening a network connection to retrieve http://www.apple.com/DTDs/PropertyList-1.0.dtd for no benefit.

Thread [Worker-18: Look up for installed JVMs] (Suspended (breakpoint at line 1051 in HttpURLConnection))	
	HttpURLConnection.connect() line: 1051	
	HttpURLConnection.getInputStream0() line: 1665	
	HttpURLConnection.getInputStream() line: 1589	
	XMLEntityManager.setupCurrentEntity(boolean, String, XMLInputSource, boolean, boolean) line: 677	
	XMLEntityManager.startEntity(boolean, String, XMLInputSource, boolean, boolean) line: 1397	
	XMLEntityManager.startDTDEntity(XMLInputSource) line: 1363	
	XMLDTDScannerImpl.setInputSource(XMLInputSource) line: 257	
	XMLDocumentScannerImpl$DTDDriver.dispatch(boolean) line: 1152	
	XMLDocumentScannerImpl$DTDDriver.next() line: 1040	
	XMLDocumentScannerImpl$PrologDriver.next() line: 943	
	XMLDocumentScannerImpl.next() line: 605	
	XMLDocumentScannerImpl(XMLDocumentFragmentScannerImpl).scanDocument(boolean) line: 542	
	XIncludeAwareParserConfiguration(XML11Configuration).parse(boolean) line: 889	
	XIncludeAwareParserConfiguration(XML11Configuration).parse(XMLInputSource) line: 825	
	DOMParser(XMLParser).parse(XMLInputSource) line: 141	
	DOMParser.parse(InputSource) line: 247	
	DocumentBuilderImpl.parse(InputSource) line: 342	
	PListParser.parseXML(InputStream) line: 104	
	PListParser.parse(InputStream) line: 67	
	MacInstalledJREs.parseJREInfo(InputStream, IProgressMonitor) line: 160	
	MacInstalledJREs.parseJREInfo(IProcess, IProgressMonitor) line: 143	
	MacInstalledJREs.getInstalledJREs(IProgressMonitor) line: 113	
	DetectVMInstallationsJob.run(IProgressMonitor) line: 67	
	Worker.run() line: 63

[...]

PR #324, opened 2023-10-13
Resolve PLIST entities to zero-length content.

nitind:
https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/323 Not doing this causes the parsing to still attempt to retrieve DTDs.

### DIFF
--- a/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/PListParser.java
+++ b/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/PListParser.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2012 IBM Corporation and others.
+ * Copyright (c) 2010, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -14,6 +14,7 @@
 package org.eclipse.jdt.internal.launching;
 
 import java.io.BufferedInputStream;
+import java.io.ByteArrayInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


iloveeclipse:
@nitind : I'm not familiar with XML parsing at all, could you please give more insights what exactly the patch is doing and why it is safe to "Resolve PLIST entities to zero-length content". I must confess I have no idea what PLIST is...

iloveeclipse:
Note: the build failures can be ignored, see https://github.com/eclipse-pde/eclipse.pde/issues/782.

@jukzi : since you've recently changed lot of XML related code in the platform, could you please check if similar change should also be applied in other places where we start XML parsing with `DocumentBuilder.parse`?

nitind:
It's a preference file for applications on macOS, in this case, content that I think is generated on the fly with no reason to use character or parameter entities. As the code is already written with the intent to not validate the contents of the XML, and there's no reason to expect any internal entities in the XML, it's simplest to drop any such requests instead of relying on parser configuration properties that may be implementation-specific.

Entity resolvers are supposed to resolve an entity into a SAX InputSource, which provides both the resolved location information and the content of what it resolved to. Returning "null" as the InputSource from a different entity resolver results in the current XML parser calling a default entity resolver as a failsafe, which then opens a network connection as it does when you don't explicitly set a resolver at all. For that reason, the proposed solution (as was implemented in WTP for reading JSP tag library and web deployment descriptor files) is to explicitly set an entity resolver that more or less drops such requests on the floor.

jukzi:
That code uses createDocumentBuilderIgnoringDOCTYPE() i.E. it should ignore Doctype i.e. ignore any DTD. the patch is not wrong though, but i don't see any benefit from it. proof me wrong with a junit

iloveeclipse:
> That code uses createDocumentBuilderIgnoringDOCTYPE() i.E. it should ignore Doctype i.e. ignore any DTD. the patch is not wrong though, but i don't see any benefit from it. proof me wrong with a junit

Have you seen the stack on the ticket?

Issue #327, opened 2023-10-18
Attribute "Without test code" is ignored in the launcher


oleosterhagen:
When adding a **project dependency** to an existing project, it is possible to set the _Without test code_ attribute, so that paths marked with _Contains test sources_ are not added to the classpath.

![grafik](https://github.com/eclipse-jdt/eclipse.jdt.debug/assets/76391368/eeb122b1-4ccd-48bb-ba4d-1b9da332df59)

This attribute is persisted as `without_test_code` in the `.classpath` file.

```
<classpathentry combineaccessrules="false" kind="src" path="/projectA">
    <attributes>
        <attribute name="without_test_code" value="true"/>
    </attributes>
</classpathentry>

```

The incremental Java compiler already takes this attribute into account, so that marked test classes from the project dependency are not accessible. But when launching a JUnit test the resolved classpath contains the output locations for the unwanted test paths. Here is an example ([example.tar.gz](https://github.com/eclipse-jdt/eclipse.jdt.debug/files/13031193/example.tar.gz)):

```
Classpath of projectA

path="main", output="bin/main"
path="test", output="bin/test", test=true


Classpath of projectB

path="test", output="bin", test=true
path="/projectA", without_test_code=true
```

When launching a test from `projectB`, classes from `/projectA/test` are accessible via reflection - but they should'nt. (The reflective access is necessary here, because in the Java editor the compiler already cannot see these classes.) It is possible to exclude all test code in the launcher (see tab _Dependencies_) - but this a different thing.

To solve this problem, the two `JavaRuntime.resolveRuntimeClasspathEntry(...)` methods ([one for launch configurations](https://github.com/eclipse-jdt/eclipse.jdt.debug/blob/399fc74f20715285613c0ced93edb5a82a5b0098/org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java#L1222), [another for projects](https://github.com/eclipse-jdt/eclipse.jdt.debug/blob/399fc74f20715285613c0ced93edb5a82a5b0098/org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java#L1470)) should evaluate the attribute `without_test_code` when resolving project dependencies.

PR #328, opened 2023-10-18
Evaluate "without_test_code" when resolving the runtime classpath #327

oleosterhagen:
When resolving the runtime classpath for project dependencies the attribute "without_test_code" has to be evaluated so that paths marked as test code from the included project do not contribute to the classpath. The contained classes or resources should not be accessible (e.g. via reflection).

## What it does

Fixes https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/327

## How to test

The example in https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/327 can be used to reproduce the problem.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
+++ b/org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
@@ -12,6 +12,7 @@
  *     IBM Corporation - initial API and implementation
  *     Frits Jalvingh - Contribution for Bug 459831 - [launching] Support attaching
  *     	external annotations to a JRE container
+ *     Ole Osterhagen - Issue 327 - Attribute "Without test code" is ignored in the launcher
  *******************************************************************************/
 package org.eclipse.jdt.launching;
 
@@ -1232,7 +1233,9 @@ public final class JavaRuntime {
 						return new IRuntimeClasspathEntry[0];
 					}
 					IClasspathAttribute[] attributes = entry.getClasspathEntry().getExtraAttributes();
-					IRuntimeClasspathEntry[] entries = resolveOutputLocations(project, entry.getClasspathProperty(), attributes, excludeTestCode);
+					boolean withoutTestCode = entry.getClasspathEntry().isWithoutTestCode();
+					IRuntimeClasspathEntry[] entries = resolveOutputLocations(project, entry.getClasspathProperty(), attributes, excludeTestCode
+							|| withoutTestCode);
 					if (entries != null) {
 						return entries;
 					}
@@ -1477,7 +1480,9 @@ public final class JavaRuntime {
 					IJavaProject jp = JavaCore.create(p);
 					if (jp != null && p.isOpen() && jp.exists()) {
 						IClasspathAttribute[] attributes = entry.getClasspathEntry().getExtraAttributes();
-						IRuntimeClasspathEntry[] entries = resolveOutputLocations(jp, entry.getClasspathProperty(), attributes, excludeTestCode);
+						boolean withoutTestCode = entry.getClasspathEntry().isWithoutTestCode();
+						IRuntimeClasspathEntry[] entries = resolveOutputLocations(jp, entry.getClasspathProperty(), attributes, excludeTestCode
+								|| withoutTestCode);
 						if (entries != null) {
 							return entries;
 						}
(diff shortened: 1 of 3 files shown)


org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
```diff
@@ -1232,7 +1233,9 @@ public static IRuntimeClasspathEntry[] resolveRuntimeClasspathEntry(IRuntimeClas
 						return new IRuntimeClasspathEntry[0];
 					}
 					IClasspathAttribute[] attributes = entry.getClasspathEntry().getExtraAttributes();
-					IRuntimeClasspathEntry[] entries = resolveOutputLocations(project, entry.getClasspathProperty(), attributes, excludeTestCode);
+					boolean withoutTestCode = entry.getClasspathEntry().isWithoutTestCode();
+					IRuntimeClasspathEntry[] entries = resolveOutputLocations(project, entry.getClasspathProperty(), attributes, excludeTestCode
+							|| withoutTestCode);
```
SarikaSinha, line 1238: If value is "No" this returns true? 
As I understood, for the value of "No" the dependency should be excluded.

SarikaSinha (changes requested):
Please explain.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
oleosterhagen: Test code should be excluded in `resolveOutputLocations` when one of these conditions is met:

1. It is explicitly requested in the *launch configuration* (This is the existing condition.):
```
boolean excludeTestCode = configuration.getAttribute(IJavaLaunchConfigurationConstants.ATTR_EXCLUDE_TEST_CODE, false);
```

2. It is stated in the classpath entry of the *project dependency* (The is the new condition introduced with this PR):
```
boolean withoutTestCode = entry.getClasspathEntry().isWithoutTestCode();
```

* When the launch configuration request exclusion of test code (case 1), test code will be excluded as before.
* When the launch configuration does not request exclusion but the current classpath entry is a project dependency and this entry is configured to not expose test code `entry.getClasspathEntry().isWithoutTestCode() `evaluates to **`true`** (case 2), so test code should also be excluded.

Currently, the `without_test_code` attribute of the classpath entry is not evaluated when resolving the runtime classpath. When resolving the classpath in the compiler, this attribute is already examined:

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/a8bfcc50993854c9584914d21b9a77427c6ebb2d/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaProject.java#L602

```
boolean nestedWithoutTestCode = excludeTestCode || entry.isWithoutTestCode();
```

oleosterhagen:
@SarikaSinha Thank you for looking at my pull request.

Does my explanation above help?

SarikaSinha:
> @SarikaSinha Thank you for looking at my pull request.
> 
> Does my explanation above help?

Yes, I was able to understand this point but this was not going well with the Help document which describes "Without Test Code". @jarthana will be putting a PR to correct it and then I can merge this.

oleosterhagen:
Thank you. I think you mean this place:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/974b10d805dc2f2502ec9bf769536f9644edee62/eclipse.platform.common/bundles/org.eclipse.jdt.doc.user/reference/ref-properties-build-path.htm#L215

_When this is set to **'No'**, test code from the referenced project is not visible when compiling test code in the current project._

This should actually corrected to **'Yes'**.

SarikaSinha (approved):

SarikaSinha:
> Thank you. I think you mean this place:
> 
> https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/974b10d805dc2f2502ec9bf769536f9644edee62/eclipse.platform.common/bundles/org.eclipse.jdt.doc.user/reference/ref-properties-build-path.htm#L215
> 
> _When this is set to **'No'**, test code from the referenced project is not visible when compiling test code in the current project._
> 
> This should actually corrected to **'Yes'**.

PR is merged now with updated document. By mistake I created merge branch, Can you update it without merge commit to be able to proceed?

oleosterhagen:
I have updated the branch.

PR #369, opened 2023-12-28
Detect Windows JVMs on startup

Bananeweizen:
## What it does
The JDK detection mechanism doesn't find much on Windows. Have it find the default installation directories of Adoptium and RedHat, both in the 64 and 32 bit installation directories. Further JDK distributors can easily be added without changing the code logic.

Also let duplicate VM names start with index 2, not 1. And finally improve the job name.

On my machine that change adds the upper 4 entries of this screenshot: 
grafik

## How to test

Install Adoptium or Redhat OpenJDK in default location and check preference page.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
+++ b/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
@@ -17,6 +17,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


jukzi:
I tested the PR and it works.
Manually i found also JDKs installed in 
`C:\Program Files\Eclipse Foundation\jdk-11.0.12.7-hotspot`
`C:\Program Files\Java\jdk1.8.0_202`
which have not been found. i don't know if that are standard paths some how, but i can not remember to have selected them manually.

I would be good to search also in the system "PATH"s, where the commandline would also find the "java" command.

It might be nice to have an information logged like 
`ILog.get().info("Detected installed JVM: " + name); //$NON-NLS-1$`

What do you think?

jukzi:
@merks how does OOmph detect JVMs?

merks:
We do similar adhoc things:

https://github.com/eclipse-oomph/oomph/blob/master/plugins/org.eclipse.oomph.jreinfo/src/org/eclipse/oomph/jreinfo/JREInfo.java

Also some native stuff to look in the system registry:

https://github.com/eclipse-oomph/oomph/blob/master/plugins/org.eclipse.oomph.jreinfo.lib/src/jreinfo.c

I think there are other more modern ways in which one could look in the system registry.  Even here you know to know for which keys to search.

Oomph as an advantage that it stores information globally for reuse across installations and workspaces...

jukzi:
@Bananeweizen 
oracles JDK installs to programm files\java\ by default
image
please add that and then this PR is good to go.

akurtakov:
What is the status here guys? Time to push or abandon this one? It's up to Windows users.

jukzi:
let's merge as is. It already is an improvement and can be further improved by subsequent changes.

PR #373, opened 2024-01-18
Use the ECJ generated compiler logs to parse compiler problems

laeubi:
It is more reliable to pare a structured file than the log output.


akurtakov:
Thanks!

PR #380, opened 2024-01-31
Jenkinsfile: align tools with jdt.ui

jukzi:


jukzi:
ignoring maven warnings - unrelated to this commit

PR #383, opened 2024-02-05
don't fail build on test error

jukzi:


iloveeclipse:
Could it be, that if we don't have any warnings and only a test fail, github will show everything green? Is therea way to test that and see how it looks like?

jukzi:
i have no fast idea how to test that, but obviously this check refers to the failed test:
image

jukzi:
i have created a PR that should not show any warning: https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/386 (by removing all Warning Plugins)

PR #384, opened 2024-02-06
remove unit-tests.yml

jukzi:


jukzi:
was an experiment:
would remove the dedicated checks:
image

image

PR #389, opened 2024-02-07
Make the snippetsupport.jar as part of the maven build

laeubi:
Currently snippetsupport.jar must be build in the ide using s ant script.

This changes the build to use maven only rebuilding the jar on each compile so class differences can be discovered automatically. The jar still needs to be checked in as it is used internally  as a resource.

## Author checklist


iloveeclipse (changes requested):
It is OK to build with maven, but why do we check in the jar to git? I would prefer not to do so, and instead keep the builder on project.

laeubi:
> It is OK to build with maven, but why do we check in the jar to git? I would prefer not to do so, and instead keep the builder on project.

Yes this would be good, but if you look at 
https://github.com/eclipse-jdt/eclipse.jdt.debug/blob/b291934c39c6f4ed33e9997911dbd0d9f0966786/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/snippeteditor/ScrapbookLauncher.java#L133

it wants this jar file, I could think of changing this part to look for the compiled classes via the classlaoder if it does not finds the jar and build one on-the-fly but that's a bit out of scope here, if you can help to resolve that issue the jar should not be required to checked in.

As an alternative one could want people to run the maven pom once before testing that feature in the IDE but I'm not sure its feasable.

iloveeclipse:
What is wrong with the external launcher that did this already but now removed?

laeubi:
> What is wrong with the external launcher that did this already but now removed?

At least in my IDE these are very disruptive as they often run after each code changes open new consoles and take some time, but in general it would be of course also possible to run a maven command there.

iloveeclipse:
> At least in my IDE these are very disruptive as they often run after each code changes open new consoles

One can setup them not to open a console.

PR #414, opened 2024-03-21
Updated bundles with comparator errors

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1923


iloveeclipse:
Build fails because of Java 22 support changes that require update from JDT core which is not yet available in a stable IBuild.

trancexpress:
Sorry, I missed that.

PR #430, opened 2024-04-16
Touching bundle after compiler change

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1979


iloveeclipse:
org.eclipse.jdt.debug.tests.ui.OpenFromClipboardTests.testQualifiedName_1 failed, but it is known to be unstable

PR #435, opened 2024-04-20
Minor clean-ups in VMInstall and its type implementations

HannesWell:
## What it does

Clean-up the code in `VMInstall` and `VMInstallType` implementations in order to simplify it.

## Author checklist

- [ ] I have thoroughly tested my changes
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #437, opened 2024-04-23
JVMs could be validated anymore

iloveeclipse:
Regression from e60dea0f97168450cf74b02745abbeea01d6417c

See https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/435


rgrunber:
Thanks for leaving a trail to follow. Got hit with this on I202422 specifically because I was trying to avoid I202423 until the API in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1386 made its way in.

Issue #301, opened 2023-09-01
[Tests] testMapLogicalStructure tests failing on gtk3


SarikaSinha:
https://download.eclipse.org/eclipse/downloads/drops4/I20230830-2200/testresults/html/org.eclipse.jdt.debug.tests_ep429I-unit-cen64-gtk3-java20_linux.gtk.x86_64_20.html
com.sun.jdi.ObjectCollectedException occurred retrieving type

org.eclipse.debug.core.DebugException: com.sun.jdi.ObjectCollectedException occurred retrieving type
at org.eclipse.jdt.internal.debug.core.model.JDIDebugElement.throwDebugException(JDIDebugElement.java:263)
at org.eclipse.jdt.internal.debug.core.model.JDIDebugElement.requestFailed(JDIDebugElement.java:202)
at org.eclipse.jdt.internal.debug.core.model.JDIDebugElement.targetRequestFailed(JDIDebugElement.java:182)
at org.eclipse.jdt.internal.debug.core.model.JDIValue.getUnderlyingType(JDIValue.java:459)
at org.eclipse.jdt.internal.debug.core.model.JDIValue.getJavaType(JDIValue.java:441)
at org.eclipse.jdt.debug.tests.variables.TestLogicalStructures.testMapLogicalStructure(TestLogicalStructures.java:109)
at java.base/java.lang.reflect.Method.invoke(Method.java:578)
at junit.framework.TestCase.runTest(TestCase.java:177)
at junit.framework.TestCase.runBare(TestCase.java:142)
at org.eclipse.jdt.debug.tests.AbstractDebugTest.runBare(AbstractDebugTest.java:2678)
at junit.framework.TestResult$1.protect(TestResult.java:122)
at junit.framework.TestResult.runProtected(TestResult.java:142)
at junit.framework.TestResult.run(TestResult.java:125)
at junit.framework.TestCase.run(TestCase.java:130)
at junit.framework.TestSuite.runTest(TestSuite.java:241)
at junit.framework.TestSuite.run(TestSuite.java:236)
at junit.framework.TestSuite.runTest(TestSuite.java:241)
at org.eclipse.jdt.debug.tests.DebugSuite$1.run(DebugSuite.java:61)
at java.base/java.lang.Thread.run(Thread.java:1623)
Caused by: com.sun.jdi.ObjectCollectedException
[...]

PR #450, opened 2024-06-14
OpenFromClipboardAction: accept whitespace after method name

jukzi:
Because VisualVm typically produces names with whitespace before parentheses which is legit in java source as well.


jukzi:
ignoring unrelated random fail TestLogicalStructures.testMapLogicalStructure https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/301

PR #458, opened 2024-07-17
Don't try to update compliance if JDT model tests are running

iloveeclipse:
The tests need predictable defaults, not based on current JVM. Check if TEST_LIB variable is known and don't update in that case. (TEST_LIB classpath variable is contributed by
org.eclipse.jdt.core.tests.model)

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/84

### DIFF
--- a/org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
+++ b/org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/JavaRuntime.java
@@ -3311,6 +3311,16 @@ public final class JavaRuntime {
 					LaunchingPlugin.log(e);
 				}
 			}
+
+			boolean testsRunning = List.of(JavaCore.getClasspathVariableNames()).contains("TEST_LIB"); //$NON-NLS-1$
+			if (testsRunning) {
+				// Don't try to update compliance if JDT model tests are running,
+				// they need predictable defaults, not based on current JVM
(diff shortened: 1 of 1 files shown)


SarikaSinha (commented):
I will prefer it handled in initializeVMs where we should set updateCompliance as false for testing condition.

iloveeclipse:
> I will prefer it handled in initializeVMs where we should set updateCompliance as false for testing condition.

Done, please check. To test if the patch works, run `JavaSearchMultipleProjectsTests` - `testMethodOccurences()` will fail without the patch (at least it does so consistently for me).

iloveeclipse:
> A related property is `-Djdt.default.test.compliance=1.8` it is set from test.xml and pom.xml.

As I've mentioned on the ticket, I'm not in favour of using system property here, because it would always need a manual update of every test launch configuration in the IDE. This would also only help during automated builds but not help contributors running tests locally.

stephan-herrmann:
> > A related property is `-Djdt.default.test.compliance=1.8` it is set from test.xml and pom.xml.
> 
> As I've mentioned on the ticket, I'm not in favour of using system property here, because it would always need a manual update of every test launch configuration in the IDE. This would also only help during automated builds but not help contributors running tests locally.

The property is already there, and at some point at least it was necessary for successful test runs (and one test even explicitly checks if it is set).

But perhaps we're lucky and after your change we can actually remove this property :)

iloveeclipse:
> The property is already there

It is not there **in the IDE**!

Issue #484, opened 2024-09-05
Tests from org.eclipse.jdt.debug.jdi.tests are not run in daily builds


jarthana:
There are bunch of tests under the suite org.eclipse.debug.jdi.tests.AutomatedSuite. These tests are run as part of the PR, but not really run as part of the daily I builds. They should be run every day too.

jukzi:
it is unclear to me what is the intended difference of
org.eclipse.debug.jdi.tests.AutomatedSuite vs
org.eclipse.debug.jdi.tests.TestAll
they look very similar.

the referenced VirtualThreadTest.testJDIPlatformThreadsOnlyFilter() fails locally to me:
```
junit.framework.AssertionFailedError: 6
	at junit.framework.TestCase.assertNull(TestCase.java:418)
```

akurtakov:
If you look at https://download.eclipse.org/eclipse/downloads/drops4/I20240909-1800/testResults.php there are 
Missing Files
org.eclipse.jdt.debug.jdi.tests_ep434I-unit-mac64-java17_macosx.cocoa.x86_64_17.xml
org.eclipse.jdt.debug.jdi.tests_ep434I-unit-macM1-java17_macosx.cocoa.aarch64_17.xml
org.eclipse.jdt.debug.jdi.tests_ep434I-unit-win32-java17_win32.win32.x86_64_17.xml
org.eclipse.jdt.debug.jdi.tests_ep434I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.xml
org.eclipse.jdt.debug.jdi.tests_ep434I-unit-cen64-gtk3-java21_linux.gtk.x86_64_21.xml
org.eclipse.jdt.debug.jdi.tests_ep434I-unit-cen64-gtk3-java22_linux.gtk.x86_64_22.xml
aka smth was missed in the enablement.

jukzi:
there is now a org.eclipse.jdt.debug.tests.log https://download.eclipse.org/eclipse/downloads/drops4/I20240909-1800/testresults/ep434I-unit-cen64-gtk3-java22_linux.gtk.x86_64_22/directorLogs/director-org.eclipse.jdt.debug.tests.log
similar to org.eclipse.jdt.debug.tests.log
[...]

PR #488, opened 2024-09-10
Add test.xml #484

jukzi:
hope that helps thats test executed in I-Builds
https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/484

### DIFF
--- a/org.eclipse.jdt.debug.jdi.tests/build.properties
+++ b/org.eclipse.jdt.debug.jdi.tests/build.properties
@@ -14,5 +14,6 @@
 source.. = tests/
 output.. = bin/
 bin.includes = about.html,\
+               test.xml,\
                META-INF/,\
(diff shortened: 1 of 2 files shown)


akurtakov (approved):
This is mandatory file but the I-builds tests run is so fragile that the only way is to merge and test.

jukzi:
@HeikoKlare any idea what is still missing?

HeikoKlare (approved):
Oh, of course, the test.xml for the plugin to test needs to exist. I went through the proposed file once and it looks sound to me (consistently defined targets and consistent with other test.xmls). But as @akurtakov mentioned, the only good way to test this is to merge and try. I think there is also a way to run the Ant build locally, but I've never tried that. So I propose to merge and check the next I-Build results.

PR #490, opened 2024-09-11
jdi tests: more logging, reduce timeout

jukzi:
jdi tests do not start on I Builds and we don't know why.


jukzi:
CI measured a good performance gain (factor 7) also for linux:
```
!ENTRY org.eclipse.core.tests.resources 1 0 2024-09-11 06:47:50.781
!MESSAGE [_testWritePerformanceBatch_(org.eclipse.core.tests.resources.IFileTest)] setUp
parallel write took:60ms
sequential write took:443ms
```
does the CI really run with so many cores?

jukzi:
ignoring unrelated 
`The method getArchive() from the type ZipEntryStorage has been deprecated since version 3.22 (removal in 2026-12 or later) and marked for removal
`

PR #502, opened 2024-09-18
Merge BETA_JAVA23 into master

jarthana:
## Author checklist


jarthana:
I believe the two failures in InstanceMainMethodsTests should go away once we have an I build with JavaSE-23. And the one in StratumTests doesn't happen in my local setup. I guess we are good to go and if they recur, we should take them up after the next I build.

SarikaSinha:
Stratum Test needs to be updated with the latest Java version for the test to pass. Will not go on it's own.

jarthana:
> Stratum Test needs to be updated with the latest Java version for the test to pass. Will not go on it's own.

OK, then we should take it up in a follow-up commit.

PR #523, opened 2024-10-01
Bugfix for #518 (Setting a breakpoint inside lambda with object).

### LINKED ISSUE #518: Setting a breakpoint inside a lamda that contains an object fails
Assume the following code

```
package example;

import java.util.Optional;
import java.util.function.Consumer;

import org.eclipse.core.databinding.observable.sideeffect.ISideEffectFactory;
import org.eclipse.jface.action.Action;
import org.eclipse.jface.databinding.swt.WidgetSideEffects;
import org.eclipse.swt.widgets.Composite;

public class NoBreakPoint {

	private Context context = new Context();

	public NoBreakPoint(Composite parent) {
		ISideEffectFactory factory = WidgetSideEffects.createFactory(parent);
		factory.create(context::getSample, sample -> new Consumer<Optional<String>>() {

			Optional<String> lastSample = Optional.empty();

			@Override
			public void accept(Optional<String> currentSample) {

				lastSample.ifPresent(System.out::println);
				currentSample.ifPresent(System.out::println);
				lastSample = currentSample;
			}
		});
	}

	private static final class Context {
		public Optional<String> getSample() {
			return Optional.of("An example"); //$NON-NLS-1$
		}
	}

	public void dummy1() {
		new Action() {

			@Override
			public void run() {
				super.run();
			}
		};
	}

}
```

- Now try to set a breakpoint in line 24 (`lastSample.ifPresent(System.out::println);`), it seems noting happens, but the Breakpoint view shows a breakpoint without a line number (in my real world code I get sett a breakpoint far away from the line i double clicked).
- Now double click on that one in the breakpoint view --> nothing happens.
- Now do the same on line 42 (`super.run();`)
- Everything works

Here is a video:

https://github.com/user-attachments/assets/112aeb33-e76d-419d-82de-975e2776e9c4

```
Eclipse SDK

Version: 2024-12 (4.34)
Build id: I20240921-1800
```

nettozahler:
## What it does
Fixes #518 

## How to test
Try to set a breakpoint inside a lambda expression with its body beginning with an anonymous class declaration.
See #518 as example. The new breakpoint should appear as usual and jumping to it from inside the "breakpoint view" must work.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.debug.tests/java8/ClassWithLambdas.java
+++ b/org.eclipse.jdt.debug.tests/java8/ClassWithLambdas.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2024 IBM Corporation and others.
+ * Copyright (c) 2024 Christian Schima.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
--- a/org.eclipse.jdt.debug.tests/java8/ClassWithLambdas.java
+++ b/org.eclipse.jdt.debug.tests/java8/ClassWithLambdas.java
@@ -1,0 +1,42 @@
+/*******************************************************************************
+ * Copyright (c) 2024 IBM Corporation and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Christian Schima - initial API and implementation
+ *******************************************************************************/
+import java.util.Optional;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+/**
+ * Test class with lambdas.
+ */
+public class ClassWithLambdas {
+	
+	private static class Factory {
+		public static <T> Factory create(Supplier<T> supplier, Consumer<T> consumer) {
+			return new Factory();
+		}
+	}
+
+	public ClassWithLambdas(String parent) {
+		Factory.create(() -> Optional.of(""), sample -> new Consumer<Optional<String>>() {
+
+			Optional<String> lastSample = Optional.empty();
+
+			@Override
+			public void accept(Optional<String> currentSample) {
+				lastSample.ifPresent(System.out::println);
+				currentSample.ifPresent(System.out::println);
+				lastSample = currentSample;
+			}
+		});
+	}
+}
--- a/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/breakpoints/TestToggleBreakpointsTarget8.java
+++ b/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/breakpoints/TestToggleBreakpointsTarget8.java
@@ -25,12 +25,8 @@ import org.eclipse.jdt.debug.tests.TestUtil;
  */
 public class TestToggleBreakpointsTarget8 extends AbstractToggleBreakpointsTarget {
 
-
-
-
 	public TestToggleBreakpointsTarget8(String name) {
 		super(name);
-		// TODO Auto-generated constructor stub
 	}
 
 	/**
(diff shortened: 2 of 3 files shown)


jukzi:
please add a junit test. see `/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/breakpoints` for related ones

nettozahler:
Hi, @jukzi, I would try to craft a PR with a JUnit test but first I need a hint to make my Eclipse SDK working again. My initial setup with OOMPH was fine and I used it for several months. Now I have updated the SDK with the "Perform setup tasks" feature. After this I now have zillions of compile errors in my SDK because all/many of the external dependencies like JSoup are missing. How can I restore/recreate all these dependencies to get rid of the compile errors? Thanks in advance.

jukzi:
When i have such problems it helps to 1. pull all repositories, 2. use the "edit" button of the eclipse-sdk target platform to update it. 3. reload the active target platform. 4. rebuild all projects. Don't know why such is not done automatically @merks

merks:
Help -> Perform Setup Tasks...  should generally update everything after pulling all the git repositories.  The *.target file should be resolved automatically by those steps, but for unknown reasons (in PDE) that doesn't always work.  :-(

nettozahler:
Thanks for the quick response with valuable hints from experienced Eclipse developers. But they have produced some big question marks on my face. As you can see I need some more detailed info. Let's go through the steps.

1. Pull all repos: "Perform Setup Tasks" does this and I can also do it manually. CHECK
2. use the "edit" button of the eclipse-sdk target platform to update it: Where is this button located? Inside Eclipse? Never seen this before.
3. reload the active target platform: How to do this?
4. rebuild all projects: Have done this zillions of times before. CHECK

jukzi:
windows/preference:
image

nettozahler:
Thanks. This procedure helped me out and resolved my compile errors.  :)

nettozahler:
Now I added a JUnit test.

org.eclipse.jdt.debug.tests/java8/ClassWithLambdas.java
```diff
@@ -0,0 +1,42 @@
+/*******************************************************************************
+ * Copyright (c) 2024 IBM Corporation and others.
```
jukzi, line 2: typically use your employer here

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/breakpoints/TestToggleBreakpointsTarget8.java
```diff
@@ -124,4 +120,26 @@ public void testInterfaceLineBreakpoint() throws Exception {
...
+	/**
+	 * Tests that a line breakpoints in an lambda expression works.
+	 */
+	public void testLineBreakpointInsideLambda() throws Exception {
+		Listener listener = new Listener();
+		IBreakpointManager manager = getBreakpointManager();
+		manager.addBreakpointListener(listener);
+		try {
+			Path path = new Path("java8/ClassWithLambdas.java");
+			final int lineNr = 35; // 0 based offset in document line numbers
```
jukzi, line 132: it is not so easy to understand where that breakpoint hits. idea: how about adding a comment in the source example "// breakpoint", and search that line here?

org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/breakpoints/ValidBreakpointLocationLocator.java
```diff
@@ -1066,10 +1066,16 @@ public boolean visit(LambdaExpression node) {
 
 						}
 					}
+
+				// Lambda body can be a block (which is handled above) or an (arbitrary) expression. So maybe the
+				// following cases are insufficient and should be replaced by handling the general
```
jukzi, line 1071: sounds odd to know what todo and not doing it ;-) could you instead of a comment suggest such an appropriate fix?

org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/breakpoints/ValidBreakpointLocationLocator.java
```diff
@@ -1066,10 +1066,16 @@ public boolean visit(LambdaExpression node) {
...
+				// org.eclipse.jdt.core.dom.Expression.
 				} else if (body instanceof LambdaExpression) {
 					body.accept(this);
 				} else if (body instanceof MethodInvocation) {
 					body.accept(this);
+				} else if (body instanceof ClassInstanceCreation) {
```
jukzi, line 1077: since the handling is the same may be use OR in the condition or even a general "instanceof Expression"

jukzi:
@nettozahler ping can you get this PR done, please?

nettozahler:
OK, let's try to get it done... I will add some comments.

org.eclipse.jdt.debug.tests/java8/ClassWithLambdas.java
jukzi: Then you can use your private name here

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/breakpoints/TestToggleBreakpointsTarget8.java
nettozahler: In general a good idea but all other existing tests here use the implemented approach so I suggest to keep it "as is". I also think that the comment "0 based offset in document line numbers" should be clear enough for developers to understand where that breakpoint hits.

org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/breakpoints/ValidBreakpointLocationLocator.java
nettozahler: Well, I am not 100% sure if the extended version of the fix would be OK. Maybe the gurus for Java grammar and AST can give a short advice here. @stephan-herrmann ?  But what is clear according to my tests is that the fix "as is" solves the described bug.

stephan-herrmann: Right, just handling the two specified alternatives Block and Expression sounds like the most consistent approach.

Seeing other visit methods freely invoking `.accept(this)` on general expressions, I see no reason why this should cause problems here, but I haven't studied class ValidBreakpointLocationLocator in more detail.

nettozahler:
Now let's try it with the revised PR.

jukzi:
@nettozahler thanks for contributing

PR #524, opened 2024-10-04
GH actions cleanup

akurtakov:
Enable dependabot to get GH workflows auto updated and drop rebase.yml as it's part of the standard GH UI now.


akurtakov:
Change can't be verified thus push directly.

PR #533, opened 2024-10-11
[Oomph-Setup] Add jdt.debug configuration setup

HannesWell:
## What it does
Add jdt.debug configuration setup.
Additionally add a styled and drag&drop-able Oomph Configuration button to the main README.

Part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2430

## Author checklist

### DIFF
--- a/org.eclipse.jdt.debug.setup/JdtDebugConfiguration.setup
+++ b/org.eclipse.jdt.debug.setup/JdtDebugConfiguration.setup
@@ -1,0 +1,88 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<setup:Configuration
+    xmi:version="2.0"
+    xmlns:xmi="http://www.omg.org/XMI"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns:setup="http://www.eclipse.org/oomph/setup/1.0"
+    label="JDT Debug">
+  <annotation
+      source="http://www.eclipse.org/oomph/setup/BrandingInfo">
+    <detail
+        key="imageURI">
+      <value>https://www.eclipse.org/downloads/images/committers.png</value>
+    </detail>
+    <detail
+        key="badgeLabel">
+      <value>JDT Debug</value>
+    </detail>
+  </annotation>
+  <installation
+      name="jdt.debug.installation"
+      label="JDT Debug Installation">
+    <setupTask
+        xsi:type="setup:VariableTask"
+        name="installation.id.default"
+        value="jdt-debug"/>
+    <productVersion
+        href="index:/org.eclipse.setup#//@productCatalogs[name='org.eclipse.applications']/@products[name='eclipse.platform.sdk']/@versions[name='latest']"/>
+    <description>The JDT Debug installation provides the latest tools needed to work with the project's source code.</description>
+  </installation>
+  <workspace
+      name="jdt.debug.workspace"
+      label="JDT Debug Workspace">
+    <setupTask
+        xsi:type="setup:VariableTask"
+        name="workspace.id.default"
+        value="jdt-debug-ws"/>
+    <setupTask
+        xsi:type="setup:CompoundTask"
+        name="User Preferences">
+      <annotation
+          source="http://www.eclipse.org/oomph/setup/UserPreferences">
+        <detail
+            key="/instance/org.eclipse.oomph.setup.ui/showToolBarContributions">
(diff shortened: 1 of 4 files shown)


HannesWell:
@jukzi or @iloveeclipse, could you please submit this as well? Thanks in advance.

org.eclipse.jdt.debug.setup/JdtDebugConfiguration.setup
```diff
@@ -0,0 +1,88 @@
...
+        name="eclipse.git.authentication.style"
+        defaultValue="anonymous"/>
+    <stream
+        href="index:/org.eclipse.setup#//@projectCatalogs[name='org.eclipse']/@projects[name='jdt']/@projects[name='debug']/@streams[name='master']"/>
+    <stream
+        href="index:/org.eclipse.setup#//@projectCatalogs[name='org.eclipse']/@projects[name='jdt']/@projects[name='debug']/@projects[name='tests']/@streams[name='master']"/>
+    <description>The JDT Debug workspace provides all the source code of the project.</description>
+  </workspace>
+  <description>
+    &lt;p>
```
jukzi, line 75: why is "<" encoded with lt but ">" is not gt encoded?

HannesWell: The Oomph-Editor is doing this. I assume it's not necessary.
It's the same for other configurations as well.

jukzi: Also in exiting configurations i can not read such fields:
image
it's illegal xml. @merks can you fix that?
I'll be AFK 2w

merks:
The only problem I see is that these PRs need to be merged:

https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/533
https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1710
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3071

That's the reason for these three broken references:

image

Could you please merge those too?

---

It's normal that only up to the first newline character is shown in the properties view.  The cell editor shows the  whole text:

image

The hover also shows it nicely formatted:

image

As does this page:

https://www.eclipse.org/setups/installer/?url=https://raw.githubusercontent.com/eclipse-jdt/eclipse.jdt/master/org.eclipse.jdt.releng/JDTConfiguration.setup&show=true

So no, a > is not invalid XML in element content.  That's why EMF does not escape them for element content:

https://github.com/eclipse-emf/org.eclipse.emf/blob/e810c433fd597d68a246cb384251214a301e871e/plugins/org.eclipse.emf.ecore.xmi/src/org/eclipse/emf/ecore/xmi/impl/XMLSaveImpl.java#L3339-L3375

But does for attribute content:

https://github.com/eclipse-emf/org.eclipse.emf/blob/e810c433fd597d68a246cb384251214a301e871e/plugins/org.eclipse.emf.ecore.xmi/src/org/eclipse/emf/ecore/xmi/impl/XMLSaveImpl.java#L3496-L3538

jukzi:
> So no, a > is not invalid XML in element content

`The right angle bracket (>) may be represented using the string " &gt; ", and MUST, [for compatibility](https://www.w3.org/TR/xml/#dt-compat), be escaped using either " &gt; " or a character reference when it appears in the string " ]]> " in content, when that string is not marking the end of a [CDATA section](https://www.w3.org/TR/xml/#dt-cdsection).
`
https://www.w3.org/TR/xml/

ok, in this case it is not invalid. while in general it may be invalid. So i would always encode it to be on the valid side, but that doesn't matter for this pr.

merks:
> ok, in this case it is not invalid. while in general it may be invalid. So i would always encode it to be on the valid side, but that doesn't matter for this pr.

In general it's never invalid in element content.  Moreover, EMF has done it this way for 25 years without anyone ever suggesting it should be changed.  If it ain't broke, don't fix it.  😬

HannesWell:
Thank you Jörg for submitting this!

The results look good and as expected:

grafik

And if you click it, the installer page looks good too:
https://www.eclipse.org/setups/installer/?url=https://raw.githubusercontent.com/eclipse-jdt/eclipse.jdt.debug/master/org.eclipse.jdt.debug.setup/JdtDebugConfiguration.setup&show=true

merks:
@jukzi 

FYI, the corner cases are handled properly by EMF:

image

Issue #559, opened 2024-11-12
[Test] testHyperLink throws NPE on MacOS in I20241111-1800 & I20241110-1800


SougandhS:
org.eclipse.jdt.debug.tests.core.LineTrackerTests.testHyperLink throws Null pointer exception
<img width="640" alt="image" src="https://github.com/user-attachments/assets/bb8ef303-fc18-4153-aabe-273636908263">

Cannot invoke "org.eclipse.debug.ui.console.IConsole.getDocument()" because "org.eclipse.jdt.debug.testplugin.ConsoleLineTracker.fConsole" is null

java.lang.NullPointerException: Cannot invoke "org.eclipse.debug.ui.console.IConsole.getDocument()" because "org.eclipse.jdt.debug.testplugin.ConsoleLineTracker.fConsole" is null
at org.eclipse.jdt.debug.testplugin.ConsoleLineTracker.getDocument(ConsoleLineTracker.java:67)
at org.eclipse.jdt.debug.tests.core.LineTrackerTests.testHyperLink(LineTrackerTests.java:184)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
at junit.framework.TestCase.runTest(TestCase.java:177)
at junit.framework.TestCase.runBare(TestCase.java:142)
at org.eclipse.jdt.debug.tests.AbstractDebugTest.runBare(AbstractDebugTest.java:2774)
at junit.framework.TestResult$1.protect(TestResult.java:122)
at junit.framework.TestResult.runProtected(TestResult.java:142)
at junit.framework.TestResult.run(TestResult.java:125)
at junit.framework.TestCase.run(TestCase.java:130)
at junit.framework.TestSuite.runTest(TestSuite.java:241)
at junit.framework.TestSuite.run(TestSuite.java:236)
at junit.framework.TestSuite.runTest(TestSuite.java:241)
at org.eclipse.jdt.debug.tests.DebugSuite$1.run(DebugSuite.java:61)
at java.base/java.lang.Thread.run(Thread.java:833)

trancexpress:
Failed again here: https://download.eclipse.org/eclipse/downloads/drops4/I20260606-1800/testresults/html/org.eclipse.jdt.debug.tests_ep441I-unit-macosx-aarch64-java21_macosx.cocoa.aarch64_21.html

```
[...]

PR #560, opened 2024-11-12
Null check for testHyperLink NPE error #559

SougandhS:
This commit adds a null check prior invoking getHyperLink method as there is a possibility for throwing NPE

Fixes : https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/559

## Author checklist

### DIFF
--- a/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/LineTrackerTests.java
+++ b/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/LineTrackerTests.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corporation and others.
+ * Copyright (c) 2000, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -181,6 +181,14 @@ public class LineTrackerTests extends AbstractDebugTest implements IConsoleLineT
 	    try {
 			ConsoleLineTracker.setDelegate(this);
 			fTarget = launchAndTerminate("ThrowsNPE");
+			long startTime = System.nanoTime();
+			long timeOut = 6000 * 1_000_000;
+			while (ConsoleLineTracker.getDocument() == null) {
+				if (System.nanoTime() - startTime > timeOut) {
+					break;
+				}
+				Thread.sleep(200);
+			}
 			getHyperlink(0, ConsoleLineTracker.getDocument());
 	    } finally {
 	        ConsoleLineTracker.setDelegate(null);


jukzi (changes requested):
But the tests looks more like it is the purpose to test that a hyperlink can be get. If that part of the test is skipped the test becomes useless. Would be better to find out why document==null and  to prevent that. May be just wait for it to become != if its a timing issue.

SougandhS:
> But the tests looks more like it is the purpose to test that a hyperlink can be get. If that part of the test is skipped the test becomes useless. Would be better to find out why document==null and to prevent that. May be just wait for it to become != if its a timing issue.

I think it might be a timing issue like you said, in local it is working.
- Will provide changes for waiting sometime.

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/LineTrackerTests.java
```diff
@@ -181,6 +181,14 @@ public void testHyperLink() throws Exception {
 	    try {
 			ConsoleLineTracker.setDelegate(this);
 			fTarget = launchAndTerminate("ThrowsNPE");
+			long startTime = System.currentTimeMillis();
```
jukzi, line 184: Please use nanoTime() instead as TimeMillis is not guaranteed to be monotonic at leap seconds.

SougandhS: okay 👍

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/LineTrackerTests.java
```diff
@@ -181,6 +181,14 @@ public void testHyperLink() throws Exception {
 	    try {
 			ConsoleLineTracker.setDelegate(this);
 			fTarget = launchAndTerminate("ThrowsNPE");
+			long startTime = System.nanoTime();
+			long timeOut = 6000;
+			while(ConsoleLineTracker.getDocument()==null) {
+				if (System.currentTimeMillis() - startTime > timeOut) {
```
jukzi, line 187: compare to System.nanoTime(); and multiply timeOut by 1_000_000

SougandhS: Okay..👍

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/LineTrackerTests.java
```diff
@@ -181,6 +181,14 @@ public void testHyperLink() throws Exception {
 	    try {
 			ConsoleLineTracker.setDelegate(this);
 			fTarget = launchAndTerminate("ThrowsNPE");
+			long startTime = System.nanoTime();
+			long timeOut = 6000 * 1_000_000;
+			while (ConsoleLineTracker.getDocument() == null) {
+				if (System.currentTimeMillis() - startTime > timeOut) {
```
jukzi, line 187: nanoTime!

SougandhS: Sorry, missed that by mistake.

jukzi (approved):
we can try that

SougandhS:
Hi @jukzi, can we commit this PR ? since its a regression fix it can go in RC1 right ?

jukzi:
As far as i understand tests can be improved at any time. However we should at least wait that the jenkins build passed if its not super urgent.

SougandhS:
Thank you @jukzi

Issue #563, opened 2024-11-15
SWT "Invalid thread access" during debug


jukzi:
```
eclipse.buildId=4.34.0.I20241113-1800
java.version=21.0.2
java.vendor=Eclipse Adoptium
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=de_DE
Command-line arguments:  -os win32 -ws win32 -arch x86_64

org.eclipse.jdt.debug
Error
Fri Nov 15 15:23:47 CET 2024
Exception occurred while notifying listener: JavaLineBreakpoint on [marker=Marker [on: /hibernate/maven/hibernate-jpamodelgen-6.6.2.Final.jar, id: 2161169, type: org.eclipse.jdt.debug.javaLineBreakpointMarker, attributes: [charEnd: 4625, charStart: 4578, lineNumber: 104, message: Line breakpoint:ProcessorSessionFactory [line: 104] - ProcessorSessionFactory(ProcessingEnvironment, Map<String, String>, Map<String, Set<String>>), org.eclipse.debug.core.enabled: true, org.eclipse.debug.core.id: org.eclipse.jdt.debug, org.eclipse.jdt.debug.core.installCount: 1, org.eclipse.jdt.debug.core.suspendPolicy: 2, org.eclipse.jdt.debug.core.typeName: org.hibernate.processor.validation.ProcessorSessionFactory, org.eclipse.jdt.debug.ui.JAVA_ELEMENT_HANDLE_ID: =hibernate/maven\/hibernate-jpamodelgen-6.6.2.Final.jar=/module=/true=/<org.hibernate.processor.validation(ProcessorSessionFactory.class[ProcessorSessionFactory, org.eclipse.jdt.internal.core.JavaModelManager.handleId: =hibernate/maven\/hibernate-jpamodelgen-6.6.2.Final.jar=/module=/true=/<org.hibernate.processor.validation(ProcessorSessionFactory.class], created: 15.11.24, 15:22]], with event: ClassPrepareEvent: 25 referenceType=org.hibernate.processor.validation.ProcessorSessionFactory$ByteBuddy$gHcoemPf

org.eclipse.swt.SWTException: Invalid thread access
	at org.eclipse.swt.SWT.error(SWT.java:4922)
	at org.eclipse.swt.SWT.error(SWT.java:4837)
	at org.eclipse.swt.SWT.error(SWT.java:4808)
[...]

PR #564, opened 2024-11-15
Move the code that acquires the modal shell into UI thread

iloveeclipse:
The change in 353005075c1ab3df53959b3deee9299cfbb7d934 replaced JDIDebugUIPlugin.getActiveWorkbenchShell() with
PlatformUI.getWorkbench().getModalDialogShellProvider().getShell(), however later one behavior is undefined in API but implementation requires the caller to be in UI thread - which was not the case of the former API.

=> Move the code that acquires the modal shell into UI thread.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/563


jukzi (approved):
tested: with this PR the Error Dialog is shown again:
image

iloveeclipse:
@trancexpress : the change is small, shouldn't add regression (I thought that on original one too :-)) and could be ideally in 4.34 RC2. Please check, so we can merge.

trancexpress:
@iloveeclipse I checked, I don't see the exception with this change. Anything else you want checked? Code-wise change looks fine from my POV.

PR #573, opened 2024-12-03
Debug tests: trying to get stable

jukzi:
* currentTimeMillis is not guaranteed to be monotonic -> use nanoTime
* added missing volatile on variables that are waited on
* uniformly sleep 1 ms during wait loops


jukzi:
ignoring unrelated `baseline and build for class have different contents`

PR #575, opened 2024-12-11
Remove repetitive code from JDIModelPresentation

gzsombor:
and using smart instanceof to remove clutter

## What it does

Eliminate the very verbose and repetitive image handling code - that looks up based on JavaDebugImages or IDebugUIConstants constants the actual ImageDescriptors and Image objects.
And eliminate a lot of redundant casts with using ' instanceof' patterns.
   
## How to test

Run it, it shouldn't change any visible behaviour. 

## Author checklist

### DIFF
--- a/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JDIModelPresentation.java
+++ b/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JDIModelPresentation.java
@@ -719,20 +719,19 @@ public class JDIModelPresentation extends LabelProvider implements IDebugModelPr
 			if (item instanceof JDIReferenceListEntryVariable){
 				return getReferenceImage();
 			}
-			if (item instanceof IJavaVariable) {
-				return getVariableImage((IAdaptable) item);
+			if (item instanceof IJavaVariable variable) {
+				return getVariableImage(variable);
 			}
-			if (item instanceof IMarker) {
-				IBreakpoint bp = getBreakpoint((IMarker)item);
-				if (bp instanceof IJavaBreakpoint) {
-					return getBreakpointImage((IJavaBreakpoint)bp);
+			if (item instanceof IMarker marker) {
+				IBreakpoint bp = getBreakpoint(marker);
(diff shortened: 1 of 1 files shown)


akurtakov:
My only concern with this PR is making methods/members private and the possibility that this might break something in the SDK (not checked). Do you think you can drop that part so we can land it in without worries ?

jukzi:
> My only concern with this PR is making methods/members private a

This PR changes "internal" packages only. So it should not break anything else. If it does we can revert.

akurtakov:
My request was exactly so a revert of the whole change can be skipped as the breakage is more likely to come from public to private change.

gzsombor:
I've reverted method privatization. What needs to be built in order to check that those methods are not used outside this repository?

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/variables/JavaVariableLabelProvider.java
```diff
@@ -55,7 +55,7 @@
  */
 public class JavaVariableLabelProvider extends VariableLabelProvider implements IPreferenceChangeListener {
 
-	public static JDIModelPresentation fLabelProvider = new JDIModelPresentation();
+	private final static JDIModelPresentation fLabelProvider = new JDIModelPresentation();
```
jukzi, line 58: still public->private

jukzi:
> What needs to be built in order to check that those methods are not used outside this repository?

You can add a separate commit for public->private and then we commit that on top o let CI build test

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/variables/JavaVariableLabelProvider.java
gzsombor: Sorry, my bad! I've pushed a fixed commit

gzsombor:
I've created a separate MR for making those method private: https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/580

jukzi:
thanks for contributing

PR #576, opened 2024-12-12
Jenkinsfile: longer history, less artifacts

jukzi:
Keep more builds to get better statistics for master and BETA branches.
But less builds for others (especially PRs).

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2653

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 pipeline {
 	options {
 		timeout(time: 40, unit: 'MINUTES')
-		buildDiscarder(logRotator(numToKeepStr: '100', artifactNumToKeepStr: '15', artifactDaysToKeepStr: '99'))
+		buildDiscarder(logRotator(numToKeepStr: (env.BRANCH_NAME == 'master' || env.BRANCH_NAME ==~ 'BETA.*') ? '100':'5', artifactNumToKeepStr: (env.BRANCH_NAME == 'master' || env.BRANCH_NAME ==~ 'BETA.*') ? '15':'2'))
 		disableConcurrentBuilds(abortPrevious: true)
 		timestamps()
 	}
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 pipeline {
 	options {
 		timeout(time: 40, unit: 'MINUTES')
-		buildDiscarder(logRotator(numToKeepStr:'15'))
+		buildDiscarder(logRotator(numToKeepStr: '100', artifactNumToKeepStr: '15', artifactDaysToKeep: '99'))
 		disableConcurrentBuilds(abortPrevious: true)
 		timestamps()
 	}
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 pipeline {
 	options {
 		timeout(time: 40, unit: 'MINUTES')
-		buildDiscarder(logRotator(numToKeepStr: '100', artifactNumToKeepStr: '15', artifactDaysToKeep: '99'))
+		buildDiscarder(logRotator(numToKeepStr: '100', artifactNumToKeepStr: '15', artifactDaysToKeepStr: '99'))
 		disableConcurrentBuilds(abortPrevious: true)
 		timestamps()
 	}


akurtakov:
@fredg02 I would like to hear from you before pushing such changes .

jukzi:
cc @fredg02

Jenkinsfile
```diff
@@ -1,7 +1,7 @@
 pipeline {
 	options {
 		timeout(time: 40, unit: 'MINUTES')
-		buildDiscarder(logRotator(numToKeepStr:'15'))
+		buildDiscarder(logRotator(numToKeepStr: '100', artifactNumToKeepStr: '15', artifactDaysToKeepStr: '99'))
```
fredg02, line 4: So far, I did not see reliable behavior with `artifactDaysToKeepStr`, so I'd recommend avoiding it completely.

In this case, the combination of values also does not make much sense.  If you only keep up to 15 builds with artifacts, and you build almost daily, artifacts will rarely get 99 days old. As discussed in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2653, going with conditional values based on the branch is more robust.

I'd start with:
```
buildDiscarder(logRotator(numToKeepStr: ('master' == env.BRANCH_NAME) ? '50' : '5', artifactNumToKeepStr: '2')
```

And in a few weeks you can check back how often you wanted to access the artifacts of builds that are older than 2 builds. :slightly_smiling_face:

jukzi: > artifacts will rarely get 99 days old

look in JDT.core PRs, nobody cleans it:
image
the implementation looks solid
https://github.com/jenkinsci/jenkins/blob/5404f14af18155d62f819279a5debddcbff794c5/core/src/main/java/hudson/tasks/LogRotator.java#L213

jukzi: > And in a few weeks you can check back how often you wanted to access the artifacts of builds that are older than 2 builds.
I know i did it in the past. Not often, but after almost every vacation because things got screwed up.

akurtakov: This is because there are still open PRs for them like https://github.com/eclipse-jdt/eclipse.jdt.core/pull/30 . If/when a PR is merged/closed it will be cleaned up.

jukzi: cancel "will be" and use "would be" - nobody in jdt.core does care.

jukzi: ok, i'd say 15 builds on master/beta, 2 builds for PRs

fredg02:
>  nobody in jdt.core does care.

That's a serious issue and should be addressed. Keeping open PRs around for too long is both frustrating for the author of the PR, but also for committers and release engineers.

Jenkinsfile
fredg02: > The I-Build logfiles to find out when and why it started to fail

Which log files exactly? The Jenkins log files (e.g. [this one](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/master/528/console)) or some other logs?

fredg02: > ok, i'd say 15 builds on master/beta, 2 builds for PRs

If you are referring to the number of builds to keep with artifacts, I still recommend a lower setting, but you may test your assumption at your convenience.

jukzi: > Which log files exactly?

https://ci.eclipse.org/releng/job/AutomatedTests/job/ep435I-unit-win32-x86_64-java17/lastSuccessfulBuild/artifact/

vogella:
> > should be addressed.
> 
> come on, getting a first PR reviewed after several years (@vogella) is kinda cool.

Not sure what this refers to but I think this is very harmful behavior and not cool.

jukzi:
sure, i tried to make a joke. its sad

jukzi:
@iloveeclipse i also want to hear your opinion about the actual values, as you had increased them.

Jenkinsfile
fredg02: Hmm. That's a different job though.

jukzi: the plan would be to apply such change to all our builds

fredg02: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/JenkinsJobs/AutomatedTests/I_unit_win32.groovy#L22

jukzi:
merging to this repo to see if it works as intended

jukzi:
looks like it works as intended: Artifact are deleted for older builds:
image

image

history and statistic became longer:
image
image

Issue #283, opened 2023-07-12
OutOfMemory in Java Indexing thread when JVM detection is enabled.


rgrunber:
First noticed in the quarkus-ls maven builds at https://github.com/redhat-developer/quarkus-ls/actions/runs/5477437137

```
Error: Exception in thread "Java indexing" java.lang.OutOfMemoryError: Java heap space
	at jdk.internal.jimage.BasicImageReader.getBufferBytes(BasicImageReader.java:356)
	at jdk.internal.jimage.BasicImageReader.getResource(BasicImageReader.java:412)
	at jdk.internal.jimage.ImageReader$SharedImageReader.getResource(ImageReader.java:569)
	at jdk.internal.jimage.ImageReader.getResource(ImageReader.java:109)
	at jdk.internal.jrtfs.SystemImage$1.getResource(SystemImage.java:69)
	at jdk.internal.jrtfs.JrtFileSystem.getFileContent(JrtFileSystem.java:250)
	at jdk.internal.jrtfs.JrtFileSystem.newByteChannel(JrtFileSystem.java:343)
	at jdk.internal.jrtfs.JrtPath.newByteChannel(JrtPath.java:702)
	at jdk.internal.jrtfs.JrtFileSystemProvider.newByteChannel(JrtFileSystemProvider.java:305)
	at java.base/java.nio.file.Files.newByteChannel(Files.java:380)
	at java.base/java.nio.file.Files.newByteChannel(Files.java:432)
	at java.base/java.nio.file.Files.readAllBytes(Files.java:3288)
	at org.eclipse.jdt.internal.compiler.util.JRTUtil.safeReadBytes(JRTUtil.java:323)
	at org.eclipse.jdt.internal.compiler.util.JrtFileSystem.lambda$1(JRTUtil.java:669)
	at org.eclipse.jdt.internal.compiler.util.JrtFileSystem$$Lambda$333/0x0000000801494460.apply(Unknown Source)
	at java.base/java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1708)
	at org.eclipse.jdt.internal.compiler.util.JrtFileSystem.getFileBytes(JRTUtil.java:667)
	at org.eclipse.jdt.internal.compiler.util.JrtFileSystem.getClassfileContent(JRTUtil.java:648)
	at org.eclipse.jdt.internal.compiler.util.JRTUtil.getClassfileContent(JRTUtil.java:268)
[...]

PR #577, opened 2024-12-15
Do not automatically detect JVMs on CI systems

laeubi:
Automatic detection on CI systems can cause a lot of trouble (e.g. memory consumption, time to read the jvms, old jvms lingering around) and is generally not useful as a CI has usually a dedicated and fixed JVM to run with.

This adds a check for the common 'CI' environment variable that is usually defined by all usual CI/CD pipelines to indicate a job is running on a CI server to skip the detection regardless of preferences configuration. If one absolutely want the feature even for CI, the system-property 'jdt.jvmdetect.envCI' can be set to an empty string.

For reference:

- Github [Default environment variables](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables)
- Jenkins [Global Variable Reference](https://ci.eclipse.org/releng/pipeline-syntax/globals#env)
- GitLab [Predefined CI/CD variables reference](https://docs.gitlab.com/ee/ci/variables/predefined_variables.html)


laeubi:
@akurtakov @iloveeclipse @mickaelistria can you review/merge?

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
```diff
@@ -48,12 +48,17 @@ public class DetectVMInstallationsJob extends Job {
...
 
+	private static final String CI_ENV_VARIABLE = System.getProperty("jdt.jvmdetect.envCI", "CI"); //$NON-NLS-1$ //$NON-NLS-2$
+
 	public DetectVMInstallationsJob() {
 		super(LaunchingMessages.lookupInstalledJVMs);
 	}
 
 	@Override
 	public IStatus run(IProgressMonitor monitor) {
+		if (!CI_ENV_VARIABLE.isEmpty() && Boolean.parseBoolean(System.getenv(CI_ENV_VARIABLE))) {
```
HannesWell, line 59: This would be a bit simpler and should handle cases where `jdt.jvmdetect.envCI` is set to an empty string or alike, by simply returning `false`.
```suggestion
		if (Boolean.getBoolean(CI_ENV_VARIABLE)) {
```

HannesWell (commented):
Sounds very good. Thank you.

mickaelistria:
> Automatic detection on CI systems can cause a lot of trouble (e.g. memory consumption, time to read the jvms, old jvms lingering around) 

Could those be fixed instead?

>  is generally not useful as a CI has usually a dedicated and fixed JVM to run with.

While this might be true for our Platform test suite, I'm skeptical this is generally true enough to get part oof the actual logic deployed to all users.
So IMO, such check would better be implemented in the test suites directly, or in some JUnit rule or whatever.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
laeubi: Get boolean returns a system property CI is an environment variable.

laeubi:
> While this might be true for our Platform test suite, I'm skeptical this is generally true enough to get part oof the actual logic deployed to all users.

I can't think about any case where I want an automatic JVM detection happen on a CI server.

mickaelistria:
> I can't think about any case where I want an automatic JVM detection happen on a CI server.

Imagine someone doing JDT plugins who just enjoy that the new JVM is discovered and used by the test cases they wrote just by adding it to their environment (eg Dockerfile, Jenkinsfile). This story seems quite interesting to me: no need to configure tests to get new encvironments available to use them; the tests also behave more like the "regular" IDE end-users have so the gap between the real scenario and the reproduce is much slimmer.

laeubi:
> Imagine someone doing JDT plugins who just enjoy that the new JVM is discovered and used by the test cases they wrote just by adding it to their environment (eg Dockerfile, Jenkinsfile). 

In this imaginary case one can set `-Djdt.jvmdetect.envCI` on their CI and get automatic detection (but decreased performances and higher memory demands).

mickaelistria:
> In this imaginary case

This is a very probable case. Before taking a decision on the default behavior, one need to consider impact on the main consumers of JDT: what makes you think that none of Spring Tools, Lombok extensions, JDT-LS... would interpret this as a regression?
Asserting that this feature is not desired on CI without investigating a bit more widely is IMO wrong, at least in term of method.

> On the other hand all the real use-cases 

One piece of advice: never claim you know all the real use-cases ;)

> one can set -Djdt.jvmdetect.envCI= on their CI and get automatic detection 

The issue is that this will be hard to discover and understand for consumers who need that.

> (but decreased performances and higher memory demands).

Can this be improved ? Is there a real reason why this VM detection operation would be so expensive?

IMO, all that effort is about putting effort on a potentially harmful or annoying workaround instead of fixing issues. Not the best way forward for anyone.

laeubi:
> This is a very probable case. Before taking a decision on the default behavior, one need to consider impact on the main consumers of JDT: what makes you think that none of Spring Tools, Lombok extensions, JDT-LS... would interpret this as a regression?

Please give concrete hints the where the explicitly rely on automatic detection of JVMs in their CI build... For me the regression is/was already added with the feature what already has wasted numerous hours of analysis and multi spreads around to disable the feature on CI.

> Asserting that this feature is not desired on CI without investigating a bit more widely is IMO wrong, at least in term of method.

Currently e.g. platform test regularly fail with OOM, @akurtakov  has fixed this for one particular case where we do not need JDT but thats not always the case.

Even @iloveeclipse has said JDT itself has to disable this for all their tests explicitly as it causes constant trouble, so how could one assert it is generally useful (or needed) for others..?!?

> Can this be improved ? Is there a real reason why this VM detection operation would be so expensive?

Yes that is why I have created this PR ;-)

There is literally no benefit on a CI for this (even though claimed otherwise), as the default JVM is already detected without this feature and for the other 0.0001% of cases where 

1. I need **different** jvms
2. I have no clue where they are and I hope they are found automatically

There is an easy workaround.

> IMO, all that effort is about putting effort on a potentially harmful or annoying workaround instead of fixing issues. 

The harmful thing is that test are using more time and more ram depending on where and how many JVMs are installed on a CI server even if none if them are referenced anywhere in the build configuration.

mickaelistria:
Again, the only thing you're talking about here are Eclipse Platform tests on Eclipse infra CI; this is a narrow use-case compared to all the cases where JDT is used. If this is an issue for Platform tests (and only for it as far as we know at the moment), then the fix should be placed somewhere around Platform tests, not in the production code.

laeubi:
>  this is a narrow use-case compared to all the cases where JDT is used

And you are claiming use-cases where none exits, this **only** disables detection of JVMs on CI servers, and I can't find any one asking for that anywhere nor that it was the intend on the initially or that is useful in any way **inside a CI build**!

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
```diff
@@ -48,12 +48,17 @@ public class DetectVMInstallationsJob extends Job {
...
+			return Status.OK_STATUS;
+		}
```
mickaelistria, line 61: Enablement logic should be placed with other similar code, in `initialize`.

mickaelistria:
> And you are claiming use-cases where none exits

One can never say something doesn't exist unless they can prove it. I'm just agnostically cautious where you try to quickly and dogmatically extrapolate one single and narrow impression as a general rule.
This feature of automatically detecting VMs even on CI is desired for JDT-LS and/or vscode-java tests. This is a critical feature for some use-cases and I suspect we have it covered by CI in those JDT consuming projects, and the proposed change here would break some tests, or at least reduce the coverage and the ability for JDT-LS and vscode-java to spot actual bugs.

laeubi:
As said for the very uncommon case where one really want to test **that** specific feature you can simply set the system property and you are done. On the other hand **every** other project (e.g. m2e, pde, and hundred others) are affected by this just because they run their test on an eclipse-platform. An even worse there builds behave differently when something is installed on the system.

mickaelistria:
So you're moving from "not any" to "imaginary" to "uncommon". As I see progress in this discussion, let's continue ;)

This feature is part of something bigger which is the story of how the IDE integrates with the environment. In the tests, we do control the environment to match the ones of users and get the tests on a user-level (most of them are integration tests). We don't want to test the detection of VM downstream (that's JDT's duty to test the feature it ships), but to rely on it for our testing, just like we want users to rely on it.
The fact that Platform tests are suffering from this feature in their tests doesn't imply all tests of all JDT consumers are suffering from it and should then be responsible for setting up workarounds of workarounds; when they actually expect that JDT will work as closely as possible as it would behave on users' machine by default (and I believe it's a sane assumption to expect that what we're testing is what we deliver to users).

laeubi:
As you said it is the duty of JDT to test the feature itself, so I assume it is working and there is no need to "test" it all over again and again on completely unrelated parts, you still have not show **where** this feature is actually used in any test.

> The fact that Platform tests are suffering from this feature in their tests doesn't imply all tests of all JDT consumers are suffering from it and should then be responsible for setting up workarounds of workarounds

The only workaround is that one constantly need to disable this feature as it creates problem on CI system in numerous ways, just because it seem useful for a small group (e.g. jdt-ls) also does not imply it is useful in general, e.g. there is actually no request for such feature over years (apart for IDE use-cases).

mickaelistria:
> I assume it is working and there is no need to "test" it all over again and again on completely unrelated parts

Exactly, we don't test it, we just use it in test.
Just like we add 1 extra call to `setup-java` in a GitHub workflow file and this new VM is automatically detected together with other ones.

> The only workaround is that one constantly need to disable this feature

People who don't want this feature can disable, that fine. People who don't have problems with it should be able to benefit from it just like end-users do; CI or not.

> also does not imply it is useful in general

Let's do basic logic: the initial "We don't want it on CI Platform => No one wants it on CI Platform" statement that drove you to build this patch is wrong. I demonstrated it wrong with 1 counter example and anyway, even without counter-example, this statement could not be proven or made reliable.
My initial statement is that there is no correlation between Platform needs and consumer needs, and that this feature *could* be useful somewhere else. This statement, because it's intentionally hypothetical, is -as weak as it is in term of certainity- undoubtly true.

Now you seem to assume that "it's not useful in general => it's useful to noone", then this is pure boolean wrong.

If you re-start the brainstorming from those assertions and their correct answer, you'll understand the harm of such change better.


Let's imagine an analogy: tomorrow, the Temurin people decide that for their own CI, they want to skip garbage collection because it slows them down, or confuse some log reading or whatever fair reason to consider it. One of its developers decide -like you suggest here- to put in the code of this final delivery that's shipped to million users such a condition `if (Boolean.getBoolean("CI")) garbageCollection = null;` then what happens: everyone one that has relied on garbage collection (even if they don't know it) will then see everything using java on CI broken.

laeubi:
I nowhere state that this is for "Platform CI", that was your assumption, I just said that **even platform** gets trouble and disable the feature **everywhere**.

> Let's imagine an analogy:

Your analogy is just the reverse, JDT has enabled a feature that no one has expected to affect their CI builds, now "million of users" are affected just because it is claimed that jdt-ls (what is not even an "official" jdt project) **might** require it (still no reference to any test that actually using/requiring/asserting that feature...)

> Just like we add 1 extra call to setup-java in a GitHub workflow file and this new VM is automatically detected together with other ones.

It does not because Github action install it to `/opt/hostedtoolcache/` what is not in the search path and could change anytime to any location not known to this check).

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
jukzi: valid request + org.eclipse.jdt.internal.launching.DetectVMInstallationsJob.initialize() already has a flag "DetectVMInstallationsJob.disabled" which can be used for this purpose. On top of that there is a setting org.eclipse.jdt.internal.launching.LaunchingPlugin.PREF_DETECT_VMS_AT_STARTUP, i don't see a value in introducing even a third option

laeubi: The main point is that this is something everyone needs to know and then needs to set ... so one probably wastes a lot of time trying to find out why Tycho build is "unstable"... then probably one is lucky enough to map it to this feature and set the option hopefully not forget it next time again.

But yes initialize is probably a better place.

jukzi:
I have not read through all the discussion. Can i summarize as followed?

Votes to change default for CI builds: 2 (Hannes, Laeubi)
Votes to keep detection in CI builds: 1 (Mickael)
As far as i know Xtext had also pain with autodetection, and so did jdt.core, quarkus https://github.com/redhat-developer/quarkus-ls/blob/e2e151096779636ef313273f31dc5a1cb81ee442/CHANGELOG.md?plain=1#L126 and even vscode https://github.com/redhat-developer/vscode-java/blob/eb145dab80da78771ef82de2a9b6ea8a8cf807b5/src/javaServerStarter.ts#L179

Sounds like CI would better opt in than opt out and i don't know a single project that would opt in

jukzi:
> > Automatic detection on CI systems can cause a lot of trouble (e.g. memory consumption, time to read the jvms, old jvms lingering around)
> 
> Could those be fixed instead?

@mickaelistria do you plan to work on that?

laeubi:
I have now adjusted the PR and slightly modified the commit message:

1. Perform the check in initialize method
2. environment variable is now only checked if a value for `DetectVMInstallationsJob.disabled` is not given, so no need for another property
3. Added some comments to clarify behavior and rationale

If one do not want autodetection on CI, simply do nothing (**off by default**). If one absolutely wants this (regardless how unlikley, uncommon, whatever), could set `-DDetectVMInstallationsJob.disabled=false` (**opt-in**).

mickaelistria:
> @mickaelistria do you plan to work on that?

If one creates a proper issue report, I would probably. But it looks like most people prefer to just disable it before investigating or even reporting...

jukzi:
https://github.com/search?q=user%3Aeclipse-platform+user%3Aeclipse-jdt+DetectVMInstallationsJob&type=issues
list 12 issues, 1 unsolved. And the other basically solved by disabling DetectVMInstallationsJob
Sounds like there had been enough chances to investigate.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
```diff
@@ -46,6 +46,20 @@
  */
 public class DetectVMInstallationsJob extends Job {
 
+	/**
+	 * CI is a common variable defined in CI/CDI servers like Jenkins, Gitlab, Github, ... to indicate it is a CI environment
+	 */
+	private static final String ENV_CI = "CI"; //$NON-NLS-1$
```
jukzi, line 52: question: is that variable set in I-Builds?

jukzi (commented):
Change looks OK for me, but i would prefer if all interested parties could agree on it.

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
laeubi: See initial PR description / commit message:

grafik

laeubi:
> list 12 issues, 1 unsolved. And the other basically solved by disabling DetectVMInstallationsJob
> Sounds like there had been enough chances to investigate.

I have especially these one in mind:

- https://github.com/eclipse-platform/eclipse.platform.ui/issues/2432
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/2609

Where here the automatic JVM detection was explicitly suspected to be (one possible) culprit:
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/2432#issuecomment-2532231115

that's also the reason I really like to get this out of the equation (but leave it to other to opt-in).

org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
jukzi: sorry, i just still don't know if our I-Builds are common CI pipelines or too much handmade.

laeubi: If you click on the Jenkins Link it will lead you to our Jenkins and showing it is all there...

jukzi: ok, i double checked https://download.eclipse.org/eclipse/downloads/drops4/I20241215-1800/testresults/consolelogs/ep435I-unit-win32-x86_64-java17_win32.win32.x86_64_17_consolelog.txt
`[echoproperties] env.CI=true`

laeubi:
> Change looks OK for me, but i would prefer if all interested parties could agree on it.

Then best is to leave an approving review to indicate this.

mickaelistria:
If you look at the issue that went the deepest into technical analysis, it's probably https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/283 , and it would look like the OOM was happening again only on Eclipse infra (with its dozens of VMs https://wiki.eclipse.org/Jenkins#JDK ). To me, the problem that one needs to fix -if related- is specific to the Eclipse CI infra (and some projects already configure the disablement of auto-detection as a workaround).
That ticket got closed because of lack of activity, which one can usually interpret as not relevant anymore. So the problem was apparently solved and for a long time, we didn't hear any complaint about VM detection. And a further topic was https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1526 , patched *after* most cases where the VM detection was considered bogus and with https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1526 might even have been improved.
The last occurrence mentioning VM detection was https://github.com/eclipse-platform/eclipse.platform.ui/issues/2432#issuecomment-2532231115 , and indeed, it would be interesting to *evaluate* the responsibility of the VM detection in a suddenly new issue about memory consumption; but this is not at all what is proposed here. What is proposed here is to -without technical material to consolidate the proposal and purely based on a series of opinions and wrong assertions- fully disable the feature by default on all CI in the world. 
This is, in my opinion, a possibly harmful change at the moment.

jukzi:
> The last occurrence mentioning VM detection was [eclipse-platform/eclipse.platform.ui#2432 (comment)](https://github.com/eclipse-platform/eclipse.platform.ui/issues/2432#issuecomment-2532231115) , and indeed, it would be interesting to _evaluate_ the responsibility of the VM detection in a suddenly new issue about memory consumption

Will you do it @mickaelistria?

I do understand both sides of frustration: 
a) having non-necessary code running of unknown impact and
b) dropping a feature without actual proof.

iloveeclipse (approved):
I was away of keyboard today, but I agree that this change makes sense for Eclipse platform and most JDT related consumers like Xtext. I'm not a friend of a "surprise" JVM detection on CI as I believe that CI builds should be predictable as much as possible to save our all time for more important work. I guess 25% of my time is spent on evaluation of "random" test failures and this change gives me personally a hope to reduce this number. So thanks @laeubi for driving this.

laeubi:
> b) dropping a feature without actual proof.

The feature is not dropped at all and if we currently ask (many) people to specify a property to _disable_ it I literally see no point in not instead asking (some) people to  specify a property to _enable_ it because:

1. they either never used it intentionally and therefore won't be affected if it is now disabled by default (and save them time+resources independent how low they are)
2. or they have used it intentionally and can easily enable it again.

Especially as it is not really guaranteed that the code finds "my" JVMs, e.g. as shown above Github installed JVMs will not be found at all, on my system I have them all under `/opt/java/...` for example and so on, I just did a quick test and hacked this path into that code and it take 6 seconds to detect 24 JVMs and floods my log with a lot of errors as some are not supported by JDT anymore. This is quite fast, but this penalty would be applied **on each start of test** that opens a workbench. Furthermore I only tested on Linux, Mac+Windows have other special code pathes and at least on Windows we know it might be slow to scan folders and start new processes.

jukzi:
One more thought i had tonight: i am also sometimes nerved if CI behaves other then running tests locally. Is there a way to also disable it for junit tests if not explicitly tested? 
Also that DetectVMInstallationsJob property could be removed from
https://github.com/eclipse-jdt/eclipse.jdt.debug/blob/d9719ca5ea3a43c23df1ab2f164e716328436883/Jenkinsfile#L28

laeubi:
> One more thought i had tonight: i am also sometimes nerved if CI behaves other then running tests locally. Is there a way to also disable it for junit tests if not explicitly tested?

That could be a different topic, I'm not aware of any such things for tests but this does not mean there is none. If you like you can of course start your eclipse with a batch script that sets CI = true env variable :-)

I think we should come to a conclusion and merge this rather sooner than later early in the dev cycle, so I'd like to ask @eclipse-jdt/eclipse-jdt-project-leads to decide here how to proceed or JDT commiters or whoever likes to decide as I think all arguments are already on the table from both sides.

mickaelistria:
There are 2 JDT committers who are OK to merge it.
I'm the only one still believe this is not a good change, but I'm not a JDT committer, so my feedback here is more informative than decisive.
Now, JDT committers can take the decision of merging if they wan, without project leads approval, that would still be fair usage of the Eclipse Development Process and I would certainly not blame the process or how it's been used if the patch get merged. I will accept the committers' decision silently; until someone comes with a complaint about it and then I will come back saying I was right 😈
I could not find which project I've worked on where we relied an auto-VM detection on CI, maybe it was changed or it's one of the experiment we did put on hold (such as eclipseide-jdtls). JDT-LS seems to use stubs VM just like JDT-Core. vscode-java seems to have its own detection mechanism; I don't know why.

jukzi:
Anybody an idea how to autodisable it for running tests locally to be locally on pair? then i would merge.

jukzi:
There is an org.eclipse.ui.internal.ide.application.JUnitTestUtil.isJunitTestRunning(), but its internal

laeubi:
@jukzi If this is merged I can try to came up with a JUnit detection but this would then only be relevant in the IDE so lets don't mix the concerns here.

PR #580, opened 2024-12-16
Make a couple of methods private around JDIModelPresentation

gzsombor:
## What it does
 Couple of methods is public, however it is not used outside of that class in JDIModelPresentation

## Author checklist

### DIFF
--- a/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JDIModelPresentation.java
+++ b/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JDIModelPresentation.java
@@ -186,7 +186,7 @@ public class JDIModelPresentation extends LabelProvider implements IDebugModelPr
 	 *            the target in which a thread is required
 	 * @return thread or <code>null</code>
 	 */
-	public static IJavaThread getEvaluationThread(IJavaDebugTarget target) {
+	private static IJavaThread getEvaluationThread(IJavaDebugTarget target) {
 		IJavaStackFrame frame = EvaluationContextManager.getEvaluationContext((IWorkbenchWindow)null);
 		IJavaThread thread = null;
 		if (frame != null) {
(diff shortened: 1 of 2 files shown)


jukzi:
LGTM, let's wait one day before merge

gzsombor:
Can you merge this now?

jukzi:
thanks for contribution

jukzi:
thanks for contribution

kysmith-csg:
Hello, it would've been nice to have a warning that this was happening, or some way to migrate. We have a debugger based on JDI that extends both of these classes and reference `getDebugImageRegistry()` as well as `fLabelProvider`.

I understand these are "internal" classes, but changing something that used to be `public` into `private` with no warning isn't very nice (especially changing it to `final`!). 

The `getDebugImageRegistry()` is fine, since it is just a convenience for `JDIDebugUIPlugin.getImageDescriptorRegistry()`, but the `fLabelProvider` is breaking. It would've been nice if you at least made some kind of accessor that could be overridden (and then not made it `final`), like `setLabelProvider()` or `getLabelProvider()`.

I'm basically going to have to hack my way around this now. Any chance for a revert for next release (or at least make them `protected` and non-`final`)?

iloveeclipse:
> I'm basically going to have to hack my way around this now. Any chance for a revert for next release (or at least make them `protected` and non-`final`)?

Sure, please provide a PR.

If you use internal non-API you should also run some tests against nightly builds so you could see the breakage before release. Or, depending on what kind of non-API you are using, provide a PR to make it proper API (e.g. some new interface etc).

kysmith-csg:
I'll work on a PR, thanks.

> Or, depending on what kind of non-API you are using, provide a PR to make it proper API (e.g. some new interface etc).

We're using basically every `org.eclipse.jdt.internal.debug.core/ui` package (an exaggeration, but we really do use a lot). I guess that's the price to pay for wanting a really deep integration with the Java debugger 🙂

Issue #203, opened 2023-03-19
Compare two strings in debug view


laeubi:
One thing I recently missed was that I have a breakpoint where it stops and I have two strings in the variable view.

It would helpful to select two String variables (of course would also be usefull for List, Map, ... ) and choose from the context menu "Compare with each other" and then get a Diff-View

SarikaSinha:
Variable View is provided by Platform debug, so the basic infrastructure to compare two elements should be provided in Platform, JDT might need to add few things to support complex Data Structure comparisons.

laeubi:
Thanks, I now created:
- https://github.com/eclipse-platform/eclipse.platform.debug/issues/121

PR #581, opened 2024-12-19
Compare multiple objects in Variable View

SougandhS:
This commit provides a new feature to compare elements of commonly used objects that implements Maps, Sets, Lists, Iterables, Queues Wrappers, non primitive  Arrays and custom Types (_eg .LinkedHashMap, ConcurrentHashMap, Hashtable , IdentityHashMap, WeakHashMap, TreeMap, List, ArrayList, Stack, Vector, LinkedList, HashSet, LinkedHashSet, TreeSet, CopyOnWriteArraySet,StringBuffer, StringBuilder etc_) from Variables view and display differences of both either in diff view or in dialog box.

Fixes https://github.com/eclipse-platform/eclipse.platform/issues/531

## How to test
->Select atleast 2 objects of same type (List, Map, Set etc) from Variables view
->Choose "Compare" from menu

image
Result<br><br>
image
<br><br><br>
image
<br><br><br>
image
<br><br><br>
image

<br><br><br>
image
<br><br><br>
image
<br><br><br>
image
<br><br><br>
image
<br><br><br>
image
<br><br><br>

image
<br><br><br>
image
<br><br><br>
image
<br><br>
image
<br><br><br>

image
<br><br><br>
image
<br><br><br>
image

## Author checklist


SougandhS:
Hi @jukzi , could you please check this PR when you have time.

jukzi:
have not looked at the sourcecode, only on the screenshots yet: 
Sounds like a usefull feature
However It is not intuitive to understand to the user if the comparison compares equality (.equals()) or identity (==) and how it is done for nested collections/arrays (deep equals?)

org.eclipse.jdt.debug.ui/plugin.properties
```diff
@@ -337,3 +337,6 @@ OpenFromClipboardAction.description = Opens a Java element or a Java stack trace
 OpenFromClipboardAction.name = Open from Clipboard
 
 VariablesView.name = Variables
+
+CompareObjects.label=Compare Objects
```
jukzi, line 341: I think mentioning "Objects" is bad. Doesn't it work for primitives? How about Compare "Values" or just "Compare" only? Will it work for results of Expressions in the expressions view too?

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparision.java
```diff
@@ -0,0 +1,654 @@
...
+	 *            IJavaValue of the selected object
+	 * @throws DebugException
+	 * @return returns actual value of the object or return combination object's name and id if extraction is not possible
+	 */
+	@SuppressWarnings("nls")
+	private String objectExtraction(IJavaValue value) throws DebugException {
+		if (value.getReferenceTypeName().equals("java.lang.String")) {
+			return value.toString();
+		}
+		if (value.getReferenceTypeName().startsWith("java.lang.StringBu")) {
```
jukzi, line 59: java.lang.StringBuffer or StringBuilder

SougandhS:
> have not looked at the sourcecode, only on the screenshots yet: Sounds like a usefull feature However It is not intuitive to understand to the user if the comparison compares equality (.equals()) or identity (==) and how it is done for nested collections/arrays (deep equals?)

it actually extracts contents in list,map,sets and objects and stores it as a string data then compares those data with equals()
suppose we have lists of		
List<String> l1 = new ArrayList<>(Arrays.asList("apple", "banana"));
List<String> l2 = new ArrayList<>(Arrays.asList("apple", "banana"));
and if we do l1.equals(l2) it results true
but if we do l1.equals(l2) for list like this (having same contents but in different order) ->
List<String> l1 = new ArrayList<>(Arrays.asList("apple", "banana"));
List<String> l2 = new ArrayList<>(Arrays.asList("banana", "apple"));
it results false

but if we now select these objects and compare from menu we get "elements contain same data"

and it wont go nested extraction for that it simply compares with the object name and id

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparision.java
SougandhS: for both actually

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparision.java
```diff
@@ -0,0 +1,654 @@
...
+		int i;
+		StringBuilder contentBuilder = new StringBuilder();
+		for (IVariable v : value.getVariables()) {
+			if (v.getName().equalsIgnoreCase("value")) {
+				for (IVariable inner : v.getValue().getVariables()) {
+					i = Integer.parseInt(inner.getValue().toString());
+					if (i == 0) {
+						break;
+					}
+					c = (char) i;
```
jukzi, line 90: this code needs explanation and tests why its valid. Why is it possible and makes sense to convert String to int to char? Think about non ascii characters is the integer a codepoint. i don't know with only looking at the code.

org.eclipse.jdt.debug.ui/plugin.properties
SougandhS: It wont work with primitives, only with variable data that are accessible from variables view. 
,will change it "Compare" like you suggested.
I haven't added this in expressions view

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparision.java
```diff
@@ -0,0 +1,654 @@
...
+	 * Extracts string value of the given string object
+	 *
+	 * @param value
+	 *            IJavaValue of the selected object
+	 * @throws DebugException
+	 * @return returns actual value of the string or return combination object's name and id if extraction is not possible
+	 */
+	@SuppressWarnings("nls")
+	private String stringObjectExtraction(IJavaValue value) throws DebugException {
+		char c;
```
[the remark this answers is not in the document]
SougandhS: okay

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparision.java
SougandhS: I will add extra comments in it

SougandhS: + Tests for StringBuilder and StringBuffer is there

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparision.java
```diff
@@ -0,0 +1,654 @@
...
+	 */
+	@SuppressWarnings("nls")
+	public Map<String, Object> compareSets(IJavaVariable selectedObject1, IJavaVariable selectedObject2) throws DebugException {
+		Map<String, Object> result = new HashMap<>();
+		result.put(SELECTED_OBJECT_1, selectedObject1.getName());
+		result.put(SELECTED_OBJECT_2, selectedObject2.getName());
+		if (selectedObject1.getValue() instanceof IJavaObject javaObject1 && selectedObject2.getValue() instanceof IJavaObject javaObject2) {
+			IJavaFieldVariable f1 = javaObject1.getField("map", false);
+			IJavaFieldVariable f2 = javaObject2.getField("map", false);
+			if (f1 == null & f2 == null && javaObject1.getReferenceTypeName().startsWith("java.util.TreeSet")) {
```
jukzi, line 116: Would be better to use  (more general) Interfaces like SortedSet instead of Implementations like TreeSet

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparision.java
```diff
@@ -0,0 +1,654 @@
...
+				result.put(IMMEDIATE_RESULT_KEY, res);
+				String failureReason = "Selected objects are different ";
+				result.put(RESULT_FAILED, failureReason);
+				return result;
+			}
+			result.put(IMMEDIATE_RESULT_KEY, res);
+			return result;
+		}
+		result.put(IMMEDIATE_RESULT_KEY, res);
+		String failed = "Selected type is not IJavaValue type";
```
jukzi, line 319: Who would see this message and when? A user could not understand the meaning.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparision.java
```diff
@@ -0,0 +1,654 @@
...
+	public Map<String, Object> stringCompare(IJavaVariable v1, IJavaVariable v2) throws DebugException {
+		Map<String, Object> result = new HashMap<>();
+		result.put(SELECTED_OBJECT_1, v1.getName());
+		result.put(SELECTED_OBJECT_2, v2.getName());
+		if (v1.getValue().getValueString().equals(v2.getValue().getValueString())) {
+			result.put(IMMEDIATE_RESULT_KEY, true);
+			return result;
+		}
+		result.put(IMMEDIATE_RESULT_KEY, false);
+		result.put(RESULT_FAILED, "Strings contains different values");
```
jukzi, line 381: Hard to understand. A String contains only characters. Maybe you mean "Strings are different"? If such message is displayed to users it should by NLS

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java
```diff
@@ -0,0 +1,445 @@
...
+			if (selectedObject2 != null && selectedObject1 != null) {
+				ObjectComparision obcp = new ObjectComparision();
+				try {
+					String referenceTypeObject1 = selectedObject1.getValue().getReferenceTypeName();
+					String referenceTypeObject2 = selectedObject2.getValue().getReferenceTypeName();
+
+					if (referenceTypeObject1.equals(referenceTypeObject2)) {
+
+						String selectionReference = selectedObject1.getReferenceTypeName();
+						if (selectionReference.equals("java.lang.String")) {
```
jukzi, line 95: can you use interface CharSequence instead of String implementation?

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java
```diff
@@ -0,0 +1,445 @@
...
+
+	/**
+	 * Checks whether the given type is of any List's references
+	 *
+	 * @param reference
+	 *            Reference of the selected object
+	 * @return returns <code>true</code> if object of list reference .<code>false</code> if object is not of List any references
+	 */
+	@SuppressWarnings("nls")
+	private boolean listLookup(String reference) {
```
jukzi, line 158: better implements interface java.util.List or java.util.Collection?

jukzi (changes requested):
All messages displayed to user (execept exceptions) should be multi language aware.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparision.java
jukzi: please also use multi char codepoints in that tests. see https://stackoverflow.com/a/9834993/9549750

SougandhS: Thanks @jukzi , I will check this and do the changes required

jukzi:
i don't get why https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/203#issuecomment-1475608485 says it should be implemented in platform but this is a jdt.debug only PR

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparision.java
SougandhS: okay, will check for SortedSet

jukzi:
> it actually extracts contents in list,map,sets

I understand what it does when i look at the code but i wonder if/how that could be expressed to the user.
And what can be expected for comparing two java.util.SortedSet that have same items but different order? Are they equal? I would hope it's not only true/false but saying they have same elements but different order. And such answer could also be a valid statement of anything that does not implements Set but also for an Array for example. While a object that only implements Set without SequencedCollection can not be compared for order. However SequencedCollection is only JDK 21, so what will be the heuristic for jdk<21?

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparision.java
jukzi: then please verify both. Maybe JDK later offers java.lang.StringBungeejumper, that should not break your code.

iloveeclipse:
If that is supposed to be implemented, it should use diff viewer support for complex types, not dialogs.

jukzi:
> for n comparisons it will be too complex

Comparing n elements is only a for-loop of comparing n-1 times 2 values each.

SougandhS:
> I understand what it does when i look at the code but i wonder if/how that could be expressed to the user. And what can be expected for comparing two java.util.SortedSet that have same items but different order? Are they equal? I would hope it's not only true/false but saying they have same elements but different order. And such answer could also be a valid statement 

Yes. Totally agree. I will provide additional msgs like you suggested.

>  While a object that only implements Set without SequencedCollection can not be compared for order. However SequencedCollection is only JDK 21, so what will be the heuristic for jdk<21?

Actually in current implementation there is a reference check provided prior comparison, so SequencedCollection will only be able to compare other SequencedCollection

SougandhS:
> If that is supposed to be implemented, it should use diff viewer support for complex types, not dialogs.

Okay

SougandhS:
I made few more changes and included a custom diff view for custom objects. now it supports n-comparisons and List,Map,Set can be campared to any of its implementations . If selected objects are only 2 then dialog box is only used if its more than 2 then diff is used.

image
<br><br><br>

image
<br><br><br>

image

<br><br><br>

image
<br><br><br>

image

<br><br><br>
image

<br><br><br>

image
<br><br><br>

for custom objects fields are compared with each other

image
image

<br><br><br>

for nested objects their identity/reference is checked
image
image


<br><br><br>

image

<br><br><br>

image
<br><br><br>

image

<br><br><br>
image

<br><br><br>
image

<br><br><br>

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparision.java
SougandhS: removed it

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparision.java
SougandhS: Yes, changed it

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java
SougandhS: from getReferenceTypeName() and javaType() only "java.lang.String" is getting returned.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparision.java
SougandhS: verified both now

jukzi (changes requested):
the license should be before the package declaration and contain "and others.". Every file should contain the license.

SougandhS:
> the license should be before the package declaration and contain "and others.". Every file should contain the license.

Added the changes now.

org.eclipse.jdt.debug.tests/testprograms/CompreObjects/CompareArrayObjects.java
```diff
@@ -0,0 +1,25 @@
...
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package CompreObjects;
```
jukzi, line 14: package name should be all low letters - for example "compare_objects_test"
https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html#:~:text=Package%20names%20are%20written%20in,example.

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/variables/CompareObjectsTest.java
```diff
@@ -0,0 +1,527 @@
...
+			result = objectComparision.stringCompare(result);
+			compareResult = (Map<String, String>) result.get(s3);
+			assertTrue("Contents are different", compareResult.get("ImmediateResult").contains("Different"));
+			result.clear();
+
+			result.put(s5, objectComparision.objectValueExtraction((IJavaValue) s5.getValue()));
+			result.put(s6, objectComparision.objectValueExtraction((IJavaValue) s6.getValue()));
+			result = objectComparision.stringCompare(result);
+			compareResult = (Map<String, String>) result.get(s5);
+			assertTrue("Contents are same regardless of reference", compareResult.get("ImmediateResult").contains("Same"));
```
jukzi, line 88: Please avoid "assertTrue"/"assertFalse", as the error message would not show the actual text.
Better use `assertThat(x, containsString("foo"));`
https://stackoverflow.com/a/1092241/9549750

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/variables/CompareObjectsTest.java
```diff
@@ -0,0 +1,527 @@
...
+		IJavaLineBreakpoint bp = createLineBreakpoint(46, typeName);
+		IJavaThread mainThread = null;
+		String resultValue;
+		Object o;
+		Map<IJavaVariable, Object> result = new HashMap<>();
+		Map<String, String> compareResult = new HashMap<>();
+		try {
+			mainThread = launchToBreakpoint(typeName);
+			int hitLine = 0;
+			assertTrue("Thread should be suspended", mainThread.isSuspended());
```
jukzi, line 125: here "assertTrue" is OK, as the target value is a boolean.

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/variables/CompareObjectsTest.java
```diff
@@ -0,0 +1,527 @@
...
+
+	@SuppressWarnings("unchecked")
+	public void testForMaps() throws Exception { // Test for Maps
+
+		String typeName = "CompreObjects.CompareMapObjects";
+		IJavaLineBreakpoint bp = createLineBreakpoint(37, typeName);
+		IJavaThread mainThread = null;
+		Map<IJavaVariable, Object> result = new HashMap<>();
+		Map<String, String> compareResult = new HashMap<>();
+		String resultKey;
```
jukzi, line 205: Delay the variable declarations to it's first initialization if the result is not used outside the code block.

org.eclipse.jdt.debug.ui/plugin.xml
```diff
@@ -1544,6 +1555,32 @@
...
+            <and>
+               <objectState
+                     name="PrimitiveVariableActionFilter"
+                     value="isPrimitive">
+               </objectState>
+               <objectState
+                     name="JavaVariableFilter"
+                     value="isInstanceRetrievalAvailable">
+               </objectState>
+
```
jukzi, line 1580: needless empty line

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/CompareElementsEditor.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.List;
+import java.util.Map;
+
+import org.eclipse.jface.resource.ImageDescriptor;
+import org.eclipse.ui.IEditorInput;
+import org.eclipse.ui.IPersistableElement;
+
+public class CompareElementsEditor implements IEditorInput {
+
+	public List<Map<String, Object>> resultsList;
```
jukzi, line 25: Please add "final" modifier, if the members are not meant to be changed after constructor

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/DebugUIMessages.properties
```diff
@@ -396,4 +396,30 @@ HeapWalkingPreferencePage_3=Instance and Reference Display Settings
...
+MapValuesSame=Values same as {0}
+DifferentDataStructures=Selected objects are of different data structures
+DialogBoxObjectsSame=Selected {0} objects are same
+DialogBoxObjectsDifferent=Selected {0} objects are different
+DialogBoxListMissing={0} dont includes : {1}
+DialogBoxListItemCount={0} items count : {1}
+DialogBoxMapSame=Objects contain same keys and values
+DialogBoxMapKeySame=Objects contain same keys but different values
+DialogBoxMapValSame=Objects contain same values but different keys
+DialogBoxMapMissingVal={0} dont includes [VALUES] : {1}
```
jukzi, line 420: my english is bad, but i think "dont includes [Values]" should be "does not include values" here

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/DebugUIMessages.properties
```diff
@@ -396,4 +396,30 @@ HeapWalkingPreferencePage_3=Instance and Reference Display Settings
...
+DialogBoxMapMissingKey={0} dont includes [KEYS] : {1}
+ComparedOnIdentity=(Compared based on identity)
+ComparisionMultiHeader=Comparison for selected objects
+ObjectsNoFields={0} objects have no accessible fields to compare
```
jukzi, line 424: i wonder why that error should ever happen - isn't it possible and always better to just call a.equals(b) to compare the objects instead of deep inspection?

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
```diff
@@ -0,0 +1,986 @@
...
+	private Set<String> keyDifferences;
+	private Set<String> valueDifferences;
+
+	private Set<String> keySimilarities;
+	private Set<String> valueSimilarities;
+
+	private Map<String, String> missingKeyData;
+	private Map<String, String> missingValData;
+
+	private Map<IJavaVariable, Object> current;
```
jukzi, line 76: I guess there are way too much members declared here. Please prefer local variables. Especially "current" is unlikely to have a sense outside the code block

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java
```diff
@@ -0,0 +1,605 @@
...
+import org.eclipse.swt.widgets.Shell;
+import org.eclipse.ui.IWorkbenchPart;
+import org.eclipse.ui.IWorkbenchWindow;
+import org.eclipse.ui.IWorkbenchWindowActionDelegate;
+import org.eclipse.ui.PartInitException;
+import org.eclipse.ui.PlatformUI;
+
+public class CompareObjectsAction extends ObjectActionDelegate implements IWorkbenchWindowActionDelegate {
+	protected IWorkbenchWindow fWindow;
+	private static final String TITLE_NAME_SUCCESS = "Object Comparision"; //$NON-NLS-1$
```
jukzi, line 44: titles visible in the UI should use Multilanguage keys

org.eclipse.jdt.debug.tests/testprograms/CompreObjects/CompareArrayObjects.java
SougandhS: will correct this

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/variables/CompareObjectsTest.java
SougandhS: okay, will replace with assertThat

org.eclipse.jdt.debug.ui/plugin.xml
SougandhS: Will remove it

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/CompareElementsEditor.java
SougandhS: Yes, will update

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/DebugUIMessages.properties
SougandhS: Your's is always better than mine, will correct this. 😅

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/DebugUIMessages.properties
SougandhS: okay, will do a reference check prior individual fields check.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
SougandhS: Okay, will make changes accordingly

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java
```diff
@@ -0,0 +1,605 @@
...
+	private static final String TITLE_NAME_FAILED = "Object Comparision Failed"; //$NON-NLS-1$
+
+	private static final String EDITOR_ID = "org.eclipse.jdt.debug.compareElementsEditor"; //$NON-NLS-1$
+
+	private static final int DISPLAY_LIMIT = 20;
+
+	private String message;
```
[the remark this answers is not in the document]
SougandhS: will set it as local

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java
SougandhS: will move this to ui message property file

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/variables/CompareObjectsTest.java
SougandhS: image

since asserThat is deprecated, I can try other alternatives

SougandhS: I will replace with assertThat for all tests

org.eclipse.jdt.debug.ui/plugin.properties
```diff
@@ -337,3 +337,6 @@ OpenFromClipboardAction.description = Opens a Java element or a Java stack trace
 OpenFromClipboardAction.name = Open from Clipboard
 
 VariablesView.name = Variables
+
+CompareObjects.label=Compare
+CompareObjects.tooltip=Compare selected elements ignoring null values
```
jukzi, line 342: what is the reason to ignore null values?

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/CompareElementsEditor.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.Map;
+
+import org.eclipse.jface.resource.ImageDescriptor;
+import org.eclipse.ui.IEditorInput;
+import org.eclipse.ui.IPersistableElement;
+
+public class CompareElementsEditor implements IEditorInput {
+
+	public final List<Map<String, Object>> resultsList;
+	public final List<String> objectsName;
```
jukzi, line 26: better "objectNames"?

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/CompareElementsEditor.java
```diff
@@ -0,0 +1,75 @@
...
+
+	public final String elementsType;
+
+	public final List<String> fieldNames;
+
+	public CompareElementsEditor(List<Map<String, Object>> resultsList, List<String> objectsName, String elementsType) {
```
jukzi, line 32: it's not common to use a plural in a prefix. resultsList -> better "result",

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/CompareElementsEditor.java
```diff
@@ -0,0 +1,75 @@
...
+import org.eclipse.jface.resource.ImageDescriptor;
+import org.eclipse.ui.IEditorInput;
+import org.eclipse.ui.IPersistableElement;
+
+public class CompareElementsEditor implements IEditorInput {
+
+	public final List<Map<String, Object>> resultsList;
+	public final List<String> objectsName;
+
+	public final String elementsType;
```
jukzi, line 28: "elementTypeNames"?

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
```diff
@@ -0,0 +1,1067 @@
...
+			if (v.getName().equalsIgnoreCase("value")) {
+				return v.getValue().getValueString();
+			}
+		}
+
+		return value.toString();
+	}
+
+	/**
+	 * Extracts byte array of ASCII values received from StringBuffer,StringBuilder and converts it into its actual character value
```
jukzi, line 80: missing space after ","

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
```diff
@@ -0,0 +1,1067 @@
...
+	 *
+	 * @param value
+	 *            IJavaValue of the selected object
+	 * @throws DebugException
+	 * @return Returns string content
+	 */
+	@SuppressWarnings("nls")
+	private String stringValueExtraction(IJavaValue value) throws DebugException {
+		int i;
```
jukzi, line 89: avoid variable declarations for not reuesd variables

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
```diff
@@ -0,0 +1,1067 @@
...
+		String message;
+		Map<IJavaVariable, Object> result = new HashMap<>();
+		for (Object selection : selections) {
+			if (selection instanceof IJavaVariable selectedObject) {
+				if (selectedObject.getValue() instanceof IJavaObject javaObject) {
+					contents = setElementsExtraction(javaObject);
+					if (contents != null) {
+						result.put(selectedObject, contents);
+					} else {
+						message = NLS.bind(DebugUIMessages.CompareObjectsFailedException, new Object[] { selectedObject });
```
jukzi, line 126: i do not understand when it would happen/ be usefull to run into such exception

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
```diff
@@ -0,0 +1,1067 @@
...
+		if (root1 == null) {
+			return null;
+		}
+		IJavaObject map1 = (IJavaObject) root1.getValue();
+		IJavaFieldVariable elements1 = map1.getField("table", false);
+		List<String> l1 = new ArrayList<>();
+		Object value;
+		if (elements1.getValue() != null) {
+			if (elements1.getValue() instanceof IJavaArray arr) {
+				if (arr != null) {
```
jukzi, line 172: an "instanceof" IJavaArray is always != null, as null is not an instanceof anything

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
```diff
@@ -0,0 +1,1067 @@
...
+	 * @throws DebugException
+	 * @return Returns a Map of KeySet and ValueSet
+	 */
+	@SuppressWarnings("nls")
+	private Map<String, Object> extractFromTreeMaps(IJavaObject elements1) throws DebugException {
+		Map<String, Object> result = new HashMap<>();
+		List<String> keySet;
+		List<String> valSet;
+		IJavaFieldVariable root1 = elements1.getField("root", false);
+		keySet = new ArrayList<>();
```
jukzi, line 201: combine initialization with declaration

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
```diff
@@ -0,0 +1,1067 @@
...
+			Set<String> set1 = setMap.get(key1);
+			Map<String, String> missingData = new HashMap<>();
+			String refType1 = key1.getValue().getReferenceTypeName();
+			properties = new HashMap<>();
+
+			for (Map.Entry<IJavaVariable, Object> entrySub : compareResults.entrySet()) {
+				Set<String> differences = new HashSet<>();
+				key2 = entrySub.getKey();
+				if (key1.equals(key2)) {
+					continue;
```
jukzi, line 680: Avoid "continue" by proper if/else if possible

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
```diff
@@ -0,0 +1,1067 @@
...
+
+	/**
+	 * Compares multiple Map
+	 *
+	 * @param compareResults
+	 *            Map containing IJavaVariable and Map contents
+	 * @throws DebugException
+	 * @return Returns a Map of comparison result for given IJavaVariable
+	 */
+	@SuppressWarnings({ "nls", "unchecked", "unlikely-arg-type" })
```
jukzi, line 728: "unlikely-arg-type" sounds serious problem

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java
```diff
@@ -0,0 +1,610 @@
...
+	 *
+	 * @param reference
+	 *            Reference of the selected object
+	 * @return returns <code>true</code> if object of list reference .<code>false</code> if object is not of List any references
+	 */
+	@SuppressWarnings("nls")
+	private boolean listLookup(String reference) {
+		if (reference.startsWith("java.util.List")) {
+			return true;
+		} else if (reference.startsWith("java.util.ArrayList")) {
```
jukzi, line 316: i really dislike specail cases for concrete implementations. better use interface like List/Collection/Iterable/Queue whatever you really need. But custom types that implement such interfaces should be handled same way

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java
```diff
@@ -0,0 +1,610 @@
...
+	/**
+	 * Displays comparison result for List objects in dialog box
+	 *
+	 * @param result
+	 *            Map containing details of comparison result
+	 * @throws DebugException
+	 */
+	@SuppressWarnings({ "nls", "unchecked" })
+	private void displayListResultsInDialogBox(Map<IJavaVariable, Object> result) throws DebugException {
+		int dispLimit = 20;
```
jukzi, line 437: "displayLimit"

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java
```diff
@@ -0,0 +1,610 @@
...
+		Map<String, Object> obj2 = (Map<String, Object>) result.get(selections.get(1));
+
+		String size1 = obj1.get(ObjectComparison.ELEMENT_SIZE).toString();
+		String size2 = obj2.get(ObjectComparison.ELEMENT_SIZE).toString();
+		if (obj1.get(ObjectComparison.IMMEDIATE_RESULT_KEY) != null || obj2.get(ObjectComparison.IMMEDIATE_RESULT_KEY) != null) {
+			outputMessage.append(obj1.get(ObjectComparison.IMMEDIATE_RESULT_KEY));
+			MessageDialog.openInformation(getShell(), DebugUIMessages.ObjectComparisonTitle_0, outputMessage.toString());
+			return;
+		}
+		String Selection1 = selections.get(0).getName().toString();
```
jukzi, line 454: variables should be low letters only -> "selection1"

org.eclipse.jdt.debug.tests/testprograms/compare/CompareListObjects.java
```diff
@@ -0,0 +1,48 @@
...
+	    linkedList1.add(12);
+	    List<Integer> linkedList2 = new LinkedList<>();
+	    linkedList2.add(12);
+	    linkedList2.add(221);
+	    
+	    List<String> Stack = new Stack<>();
+		Stack.add("Cherry");
+		Stack.add("Banana");
+		Stack.add("Apple");
+        List<String> ArrayList = new ArrayList<>();
```
jukzi, line 34: unaligned leading spaces. please autoformat the code.

jukzi (changes requested):
Wow, that code became really long. Unfortunatly it is hard to read when the declarations are not at the initializing code.

org.eclipse.jdt.debug.ui/plugin.properties
SougandhS: image <br>
There was this special case with HashSets and some Map implementations which always return a fixed size array having actual values followed by null values. so instead of doing workarounds I simply ignored nulls in every collection as it was simply just Null vals. Now that you mentioned I will do a workaround to retain the actual Null values present in the Collection

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java
SougandhS: I agree, But when using `IJavaVariable.getValue().getReferenceTypeName()` and `IJavaVariable.getReferenceTypeName()`, I was only able to retrieve the exact values provided in the source. I will it check again

SougandhS:
> Wow, that code became really long. Unfortunatly it is hard to read when the declarations are not at the initializing code.

Will clean it further more

org.eclipse.jdt.debug.ui/plugin.properties
SougandhS: Modified code to support Null comparisons

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/CompareElementsEditor.java
SougandhS: Refactored

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/CompareElementsEditor.java
SougandhS: Refactored now

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/CompareElementsEditor.java
SougandhS: Refactored now

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/DebugUIMessages.properties
```diff
@@ -396,4 +396,36 @@ HeapWalkingPreferencePage_3=Instance and Reference Display Settings
...
+DialogBoxObjectsDifferent=Selected {0} objects are different
+DialogBoxListMissing={0} does not includes : {1}
+DialogBoxListItemCount={0} items count : {1}
+DialogBoxMapSame=Objects contain same keys and values
+DialogBoxMapKeySame=Objects contain same keys but different values
+DialogBoxMapValSame=Objects contain same values but different keys
+DialogBoxMapMissingVal={0} does not includes [VALUES] : {1}
+DialogBoxMapMissingKey={0} does not includes [KEYS] : {1}
+ComparisionMultiHeader=Comparison for selected objects
+ObjectComparisonTitle_0=Object Comparison
```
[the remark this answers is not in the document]
SougandhS: Refactored now

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
SougandhS: Added missing space

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
SougandhS: Modified accordingly

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
SougandhS: done

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
SougandhS: Removed `continue` in possible places

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java
SougandhS: Refactored now

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java
SougandhS: refactored now

org.eclipse.jdt.debug.tests/testprograms/compare/CompareListObjects.java
SougandhS: Done

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/DebugUIMessages.properties
```diff
@@ -396,4 +396,35 @@ HeapWalkingPreferencePage_3=Instance and Reference Display Settings
...
+DialogBoxMapValSame=Objects contain same values but different keys
+DialogBoxMapMissingVal={0} does not includes [VALUES] : {1}
+DialogBoxMapMissingKey={0} does not includes [KEYS] : {1}
+ComparisionMultiHeader=Comparison for selected objects
+ObjectComparisonTitle_long=Object Comparison
+ObjectComparisonTitle_failed=Object Comparison Failed
+ObjectComparisonTitle_short=Comparison 
+CompareObjectsReference=Reference Mismatch
+ObjectsReferenceDifferent=Different in {0}
+ObjectsReferenceSameAndDifferent=Same in {0} and Different in {1}
```
jukzi, line 429:
```suggestion
ObjectsReferenceSameAndDifferent=Same in {0} and different in {1}
```

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
```diff
@@ -0,0 +1,970 @@
...
+	/**
+	 * Extracts all fields and its values for custom objects if fields are available else throw exception
+	 *
+	 * @param javaObject1
+	 *            IJavaObject of selected object
+	 * @throws DebugException
+	 * @return Returns a Map of IJavaVariable and field value pair
+	 */
+	public Map<String, String> customObjectValueExtraction(IJavaObject javaObject1) throws DebugException {
+		String content;
```
jukzi, line 432: please put this declaration in the block (for loop) where it is actually used

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
```diff
@@ -0,0 +1,970 @@
...
+	/**
+	 * Extracts individual array elements
+	 *
+	 * @param selectedObject1
+	 *            IJavaVariable of selected object
+	 * @throws DebugException
+	 * @return Returns a List of array elements
+	 */
+	public List<String> arrayElementsExtraction(IJavaVariable selectedObject1) throws DebugException {
+		String val1;
```
jukzi, line 496: declare where  initialized

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
```diff
@@ -0,0 +1,970 @@
...
+	 * Compares multiple strings
+	 *
+	 * @param compareResults
+	 *            Map containing IJavaVariable and its string content
+	 * @throws DebugException
+	 * @return Returns a Map of comparison result for given IJavaVariable
+	 */
+	public Map<IJavaVariable, Object> stringCompare(Map<IJavaVariable, Object> compareResults) throws DebugException {
+		Map<IJavaVariable, Object> result = new HashMap<>();
+		String val1;
```
jukzi, line 536: still lot of early declarations in this file

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
jukzi: Please try to understand the unlikely-arg-type warning and get rid of it.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java
```diff
@@ -0,0 +1,609 @@
...
+	 *            List of elements object have
+	 * @return returns a modified list of elements in the object
+	 */
+	@SuppressWarnings("nls")
+	private String printListContents(String Selection, List<String> list) {
+		int displayLimit = 30;
+		StringBuffer content = new StringBuffer();
+		int i = 0;
+		content.append("\n" + Selection + " does not include [");
+		for (String s : list) {
```
jukzi, line 574: such for loop with limit can be simpler written like `list.stream().limit(DISPLAY_LIMIT).collectors(Collect.joining(","))`.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java
```diff
@@ -0,0 +1,609 @@
...
+	 */
+	private void displayComparisonFailed(String message) {
+		MessageDialog.openError(getShell(), DebugUIMessages.ObjectComparisonTitle_failed, message);
+	}
+
+	protected Shell getShell() {
+		if (fWindow != null) {
+			return fWindow.getShell();
+		}
+		if (getWorkbenchWindow() != null) {
```
jukzi, line 604: since it is not obvious that successive calls of getWokrbenchWindow() return the same value it is better to store the valuee in a local copy or use the instanceof pattern matching

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/DebugUIMessages.properties
SougandhS: renamed now

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
SougandhS: Done

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
SougandhS: Now refactored all

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java
SougandhS: okay

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java
SougandhS: extracted into local

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/ObjectComparison.java
SougandhS: Fixed unlikely-arg-type warning now

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java
SougandhS: Hi @jukzi 
I tried by extracting interfaces 
image
but the issue is for some implementations like Enums, ConcurrentSkipListSet etc I was unable to extract stored value and if a custom type implements any and store data in any other field or another way, it wont be able to find the exact field that stores the value so it can break. so can we go with concrete implementations now ?

jukzi: I do not understand the issue that you have, especially how Enums are involved.
In the variables view it is also possible to show Logical Structures which uses the interface Collection:
image
Given that there are thousends of custom List implementations it would be very sad if those could not be used:
image

SougandhS: util.Collections cant be used, since it is used in both Set and Lists, 
I'll try again

SougandhS: Will try other extraction methods

SougandhS: Hi @jukzi 
Now I have implemented another approach for extraction (Which I should've done in the first place) and it works based on Implementations like you suggested 👍 , I have checked with most concretes with this implementations and it works

jukzi (commented):
Looks much better now, i don't know when i would have time to test it. maybe meanwhile some one else wants to take a look? @iloveeclipse ?

jukzi:
i tested with

```
public class A {
		Set<String> s1 = java.util.concurrent.ConcurrentHashMap.newKeySet();
		s1.add("1");
		Set<String> s2 = Set.of("2");
		Set<String> s11 = Set.of("1");
		List<String> l1 = List.of("1");
		System.out.println(s1.equals(s11)); // true
		System.out.println(s1.equals(s2)); //false
		System.out.println(s1.equals(l1)); //false
	}
}

```
This message seems to be inconsitent:
image
either name both objects i.e. "s1 equals s11" or none, i.e. "Objects are equal"

For the record, the other messages shown are:
image
image

I dislike the "error" icon. its not an error when objects are of different type. its just a result of comparison:
image

jukzi:
I also tried to compare 3 objects.
```
		Set<String> s2 = Set.of("2","3");
		Set<String> s11 = Set.of("1","3");
		Set<String> s3 = Set.of("4","3");

```
Results in:
image
looks ugly: "SL No." is unclear. dark red backgroundcolor makes a bad contrast,. The "Missing Elements" column is hard to understand. The simple Listing in the Variables View is much better to understand:
image

jukzi:
For Lists:
```
		List<String> l1 = new ArrayList<>();
		l1.add("1");
		l1.add("3");
		List<String> l2 = List.of("3","1");
```
image
instead of "Object" better say "Collections contain same elements, but in different order" where "Collection" should be the Interface that you actually used for comparison (i did not check which Interface the implementation uses)

SougandhS:
Made changes as suggested<br>

>either name both objects i.e. "s1 equals s11" or none, i.e. "Objects are equal"

>instead of "Object" better say "Collections contain same elements, but in different order" where "Collection" should be the Interface that you actually used for comparison


image
<br>
image<br>
image<br><br>


>I dislike the "error" icon. its not an error when objects are of different type. its just a result of comparison:
<br>
image
<br><br>

>I also tried to compare 3 objects.

dark red with black font was ugly (mine by default it was showing with white), so now made it always with white font

image

<br><br>

>The simple Listing in the Variables View is much better to understand

The summarised comparison results provides a focused view of exactly what is missing from each list. Instead of overwhelming the user with all the list data, it highlights the key differences, making it easier to spot what needs attention. Lets say we have 50+ list with 50+ contents, then it would be very difficult to find the missing values from simple listing of all, user may need to write additional code in debug shell just to find the missing ones alone.

This feature is particularly useful for large datasets, as it helps avoid information overload and makes it easier to focus on the important differences between the collections.

>instead of "Object" better say "Collections contain same elements, but in different order" where "Collection" should be the Interface that you actually used for comparison (i did not check which Interface the implementation uses)

image

laeubi:
Many thanks for working on this, this looks very useful!

jukzi:
There is a build error now, which needs you to updated the bundle to java21 (see https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2739)
```
00:07:24.297  [ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:4.0.11-SNAPSHOT:compile (default-compile) on project org.eclipse.jdt.debug.ui: Compilation failure: Compilation failure: 
00:07:24.297  [ERROR] /home/jenkins/agent/workspace/eclipse.jdt.debug-github_PR-581/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java:[493] 
00:07:24.297  [ERROR] 	if (getWorkbenchWindow() instanceof IWorkbenchWindow win) {
00:07:24.297  [ERROR] 	    ^^^^^^^^^^^^^^^^^^^^
00:07:24.297  [ERROR] Expression type cannot be a subtype of the Pattern type
```

SougandhS:
> There is a build error now, which needs you to updated the bundle to java21 (see [eclipse-platform/eclipse.platform.releng.aggregator#2739](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2739))

fixed now 👍

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/actions/CompareObjectsAction.java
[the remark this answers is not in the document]
jukzi: please make "message" local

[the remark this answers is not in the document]
SougandhS: done

SougandhS:
I have made minor changes in messages and updated the PR description with latest screenshots 👍

jukzi:
I merge it so that more people can easily test it. I think its not perfect but already adds value and could be improved further later. I suggest to add a N&N entry.

SougandhS:
> I merge it so that more people can easily test it. I think its not perfect but already adds value and could be improved further later. 

Thank you @jukzi  , I agree with your words, since its a brand new feature on java IDEs

>I suggest to add a N&N entry.

Will include this

Issue #584, opened 2025-01-08
"OpenFromClipboardTests does not exist"


jukzi:
https://download.eclipse.org/eclipse/downloads/drops4/I20250107-1800/testresults/ep435I-unit-linux-x86_64-java21_linux.gtk.x86_64_21/org.eclipse.jdt.debug.tests.AutomatedSuite.txt

contains entries that i do not understand:

```
!ENTRY org.eclipse.jdt.debug.tests 1 0 2025-01-08 01:01:06.064
!MESSAGE [testMultipleInfixEval10] setUp

!ENTRY org.eclipse.jdt.launching 4 4 2025-01-08 01:01:06.741
!MESSAGE Advanced source lookup task failed
!SUBENTRY 1 org.eclipse.jdt.core 4 969 2025-01-08 01:01:06.741
!MESSAGE OpenFromClipboardTests does not exist
!SUBENTRY 1 org.eclipse.jdt.core 4 969 2025-01-08 01:01:06.741
!MESSAGE OpenFromClipboardTests does not exist
!SUBENTRY 1 org.eclipse.jdt.core 4 969 2025-01-08 01:01:06.741
!MESSAGE OpenFromClipboardTests does not exist
!SUBENTRY 1 org.eclipse.jdt.core 4 969 2025-01-08 01:01:06.741
!MESSAGE OpenFromClipboardTests does not exist
!SUBENTRY 1 org.eclipse.jdt.core 4 969 2025-01-08 01:01:06.741
!MESSAGE OpenFromClipboardTests does not exist

!ENTRY org.eclipse.jdt.debug.tests 1 0 2025-01-08 01:01:06.844
!MESSAGE [testMultipleInfixEval10] tearDown
```

However OpenFromClipboardTests  seems to be executed.
https://download.eclipse.org/eclipse/downloads/drops4/I20250107-1800/testresults/xml/org.eclipse.jdt.debug.tests_ep435I-unit-macosx-aarch64-java21_macosx.cocoa.aarch64_21.xml

![image](https://github.com/user-attachments/assets/f185cb27-1d8a-4dbd-9a39-a20ea7ffc3b4)

PR #585, opened 2025-01-08
[test] beautify logfile output

jukzi:
Also explicitly log the testclass to identify the cause of "OpenFromClipboardTests does not exist"

https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/584


jukzi:
before:
image
after:
image

akurtakov:
I really wonder how is this output helpful at all?

iloveeclipse:
> I really wonder how is this output helpful at all?

If tests hang or report errors without stack etc you will see who was that

jukzi:
when there is some exception/logging during a test it is clear to which tests it belongs. Especially if a exception does happen in another thread that does not have the test in the stacktrace

Issue #592, opened 2025-01-14
Primitive array's signature is not shown in variables view


SougandhS:
Compared to other arrays primitive arrays type is not shown in variable view's value column <br>
**Current** - 
<img width="716" alt="image" src="https://github.com/user-attachments/assets/fe0e9b27-7bbf-441e-9902-fae452082435" />
<br>
**Proposed** -
<img width="404" alt="image" src="https://github.com/user-attachments/assets/97fb0e82-3299-4d44-96ca-f9718e2a997c" />

jukzi:
And for lambdas it seems to be a nonreadable hash:
![image](https://github.com/user-attachments/assets/6f310b6d-dfde-4369-bba0-965cdaf47d79)

PR #595, opened 2025-01-15
Primitive arrays and Lambdas type on variables view

SougandhS:
https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/592

This commit provides fix for not showing primitive arrays and lambdas correctly in variables view.

Fixes : https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/592

Before fix - 
<br>
image
<br>
After - 
image

## Author checklist

### DIFF
--- a/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JDIModelPresentation.java
+++ b/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JDIModelPresentation.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2024 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JDIModelPresentation.java
```diff
@@ -1917,11 +1924,20 @@ public String removeQualifierFromGenericName(String qualifiedName) {
 	/**
 	 * Return the simple name from a qualified name (non-generic)
 	 */
+	@SuppressWarnings("nls")
```
jukzi, line 1927: Will this work for ShowQualifiedNames? While it is unlikely, technical any class could be named "$Lambda" so it would be better if somehow this could be solved by not using text replacement only.

SougandhS: with ShowQualifiedNames enabled it will show its original value eg. `$$Lambda.0x000000f001007608` . Unfortunately vm only gives this as `IValue` for every lambdas compared to other objects. so only went with text replacement.

jukzi (approved):

SougandhS:
Thank you @jukzi

PR #597, opened 2025-01-19
Update tests relating to default root modules as of JEP 261

stephan-herrmann:
Adjusting to changes from
	https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2786

+ move to new method with augmented signature
+ update set of modules
  - untainted default root modules
  - computed lists as used for --limit-modules
+ remove expected values for versions that can no longer run this test
+ add a new test that simulates Java 9 semantics using --release
  + fix project setup to actually set the promised compliance


stephan-herrmann:
I admit that this test class is not beautiful. There have been so many changes in modules contained by various JDKs and now the change from https://bugs.openjdk.org/browse/JDK-8205169 stirs up a lot of dust.

Also, at some point in the past we used to challenge different JDKs by _running the test on_ different JDKs. Now that we need 21+ to run the test, the only way to come close to previous versions is by using `--release`, but even that differs from the original JDK 9. Keeping track of changes in code comments isn't very helpful anymore either.

Hence, this PR should be seen as a best effort, rather than a 100% solution. I'm open to ideas for improving the test, but I'm running out of ideas.

PR #604, opened 2025-01-24
Restore Java 17 BREE for org.eclipse.jdt.debug

laeubi:
This is currently only needed to add an `@Override` annotation in one method but this can also be overridden without explicit annotation.

As jdt.debug is also a dependency for api tools, this currently prevents running latest API tools on java 17 as it fails with:

```
[INFO] --- tycho-apitools:4.0.11-SNAPSHOT:verify (verify) @ org.eclipse.osgi ---
[INFO] Resolve API baseline for org.eclipse.platform:org.eclipse.osgi:eclipse-plugin:3.23.0-SNAPSHOT with linux/gtk/x86_64
Error:  org.eclipse.jdt.debug 3.22.0.v20250116-1056 requires Execution Environment that matches (&(osgi.ee=JavaSE)(version=21)) but the current resolution context uses [a.jre.javase 17.0.0]
[INFO] {osgi.os=linux, osgi.ws=gtk, org.eclipse.update.install.features=true, osgi.arch=x86_64, org.eclipse.update.install.sources=true}
Error:  Cannot resolve project dependencies:
Error:    Software being installed: org.eclipse.pde.api.tools 1.3.700.v20250117-1154
Error:    Missing requirement: org.eclipse.jdt.debug 3.22.0.v20250116-1056 requires 'osgi.ee; (&(osgi.ee=JavaSE)(version=21))' but it could not be found
Error:    Cannot satisfy dependency: org.eclipse.jdt.launching 3.23.200.v20250110-0740 depends on: osgi.bundle; org.eclipse.jdt.debug [3.21.0,4.0.0)
Error:    Cannot satisfy dependency: org.eclipse.pde.api.tools 1.3.700.v20250117-1154 depends on: osgi.bundle; org.eclipse.jdt.launching [3.21.0,4.0.0)
```


jukzi:
> As jdt.debug is also a dependency for api tools, this currently prevents running latest API tools on java 17 as it fails with:

Would be ok as a workaround, But why don't we upgrade that api tools run to 21 instead? we won't keep debug on 17 forever

laeubi:
> Would be ok as a workaround, But why don't we upgrade that api tools run to 21 instead? we won't keep debug on 17 forever

Because then it still wont run on Java 17 anymore... So this wont help anything regarding "this currently prevents running latest API tools on java 17".

> We won't keep debug on 17 forever

Nothing is forever, I just see we can keep it _now_ at 17 without loosing anything.

jukzi:
> this currently prevents running latest API tools on java 17"

Why is that required? Only shifts problem into the future. @akurtakov wdyt?

laeubi:
> > this currently prevents running latest API tools on java 17"
> 
> Why is that required? Only shifts problem into the future. @akurtakov wdyt?

Because people call for help here instead of fixing the problem:

- https://github.com/eclipse-equinox/equinox/pull/765#issuecomment-2612510134

merks:
I hope this does not spiral down the dark path.  Please try to be respectful of differences of opinion, focus, and priorities.

I expect that the community using Tycho 4.0.x is best served if it can simply continue to work without disruption.  This is not a hard requirement but a very nice-to-have.  If it's not too great a burden, that seems a reasonable request.

jukzi:
My only concern is that this will not be the last time/only plugin that is requested to stay on 17 while we already decided to move to 21. I still think it's the best (easiest) to just bulk move every bundle to 21.

merks:
Yes I empathize with the alternative view.  It’s arguably easy and consistent to mass move.  Life is a trade off though with costs and benefits to be balanced.  Thanks for taking that into consideration.

laeubi:
Thanks for the merge, API tools is running again:

https://ci.eclipse.org/releng/job/equinox/job/PR-765/2/console

Basically I'm all for upgrading, but as I mentioned elsewhere "upgrading" do not mean dumb search+replace of some Strings, and especially here it shows a ripple effect so if it is done and:

1. The one upgrading also fixes the new warnings instead of let them grow
2. Investigate and help fixing side effects (e.g. here upgrading all test and infrastructure for Equinox to Java 21) in platform and its tools (e.g. SecurityManager)
3. Help "other" projects if they are suddenly forced to upgrade their setups

I would have no concerns about it. It is just that at the moment (and especially with API tools!) things have settled and are working and I'm not eager to "just" upgrade and have a tons of complains than there is "only" this little thing to upgrade and soon you have spend months in clean up behind for little gain.

With the next release of Eclipse, we can also again "fix" the dependencies and then Tycho/Builds would not be directly affected by such changes (but might no longer be able to upgrade to latest bug fixes/features).

PR #612, opened 2025-02-03
Default formatter for Exception objects #125

SougandhS:
This commit modifies the default detail formatter of Exception objects by showing Exception class name and its stack trace by default if no specific detail formatter configured for Exception objects

Fixes : https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/125
<br>
Before <br>
image
<br>
After <br>
image <br>
image

## Author checklist

### DIFF
--- a/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JavaDetailFormattersManager.java
+++ b/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JavaDetailFormattersManager.java
@@ -382,6 +382,7 @@ public class JavaDetailFormattersManager implements IPropertyChangeListener, IDe
 	 * @return the snippet for the given class / super class
 	 * @throws DebugException if there is a problem computing the snippet
 	 */
+	@SuppressWarnings("nls")
 	private String getDetailFormatterSuperClass(IJavaClassType type) throws DebugException {
 		if (type == null) {
 			return null;
@@ -389,6 +390,15 @@ public class JavaDetailFormattersManager implements IPropertyChangeListener, IDe
 		DetailFormatter detailFormatter= fDetailFormattersMap.get(type.getName());
 		if (detailFormatter != null && detailFormatter.isEnabled()) {
 			return detailFormatter.getSnippet();
+		}
+		if ((detailFormatter == null || !detailFormatter.isEnabled()) && type.getName().equals("java.lang.Throwable")) {
+			String snippet = """
+					java.io.ByteArrayOutputStream bout = new java.io.ByteArrayOutputStream();
+					java.io.PrintStream ps = new java.io.PrintStream(bout);
+					this.printStackTrace(ps);
+					return bout.toString();
+					""";
+			return snippet;
 		}
 		return getDetailFormatterSuperClass(type.getSuperclass());
 	}
--- a/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/AbstractDebugTest.java
+++ b/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/AbstractDebugTest.java
@@ -409,6 +409,7 @@ public abstract class AbstractDebugTest extends TestCase implements  IEvaluation
 				cfgs.add(createLaunchConfiguration(jp, "a.b.c.GenericMethodEntryTest"));
 				cfgs.add(createLaunchConfiguration(jp, "org.eclipse.debug.tests.targets.HcrClass", true));
 				cfgs.add(createLaunchConfiguration(jp, "a.b.c.Bug570988"));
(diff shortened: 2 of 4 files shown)


SougandhS:
Hi @jukzi , could you please check this PR when you have time.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JavaDetailFormattersManager.java
```diff
@@ -390,6 +391,15 @@ private String getDetailFormatterSuperClass(IJavaClassType type) throws DebugExc
 		if (detailFormatter != null && detailFormatter.isEnabled()) {
 			return detailFormatter.getSnippet();
 		}
+		if ((detailFormatter == null || !detailFormatter.isEnabled()) && type.getName().equals("java.lang.Throwable")) {
+			StringBuffer buf = new StringBuffer();
```
jukzi, line 395: Can we use a text block here instead of StringBuffer? if not then use StringBuilder instead of Buffer

jukzi:
Looks like a cool feature i really missed but did not know it was possible to implement that ease, thanks.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JavaDetailFormattersManager.java
SougandhS: Will do with text block 👍

jukzi:
I tested - Unfortunately this PR removes the Exception message. Could you please keep it?
image

image
Maybe it would also be nice to keep the Format of a Typcial stacktrace i.e. having "   at" in front?

looking at the code i guess you need also support exceptioncause and supressed Exceptions, but maybe you will use java.lang.Throwable.printStackTrace(PrintStream) see
https://stackoverflow.com/a/1149712/9549750

SougandhS:
Will check this

jukzi:
image

SougandhS:
@jukzi  How about this one ? <br>
image

jukzi:
compare it to:
```
import java.io.PrintWriter;
import java.io.StringWriter;

public class Test {

		RuntimeException reason = new RuntimeException("reason");
		RuntimeException runtimeException = new RuntimeException("hello",reason);
		runtimeException.addSuppressed(new IllegalArgumentException("argument"));
		StringWriter sw = new StringWriter();
		PrintWriter pw = new PrintWriter(sw);
		runtimeException.printStackTrace(pw);
		String sStackTrace = sw.toString(); // stack trace as a string
		System.out.println(sStackTrace);
		throw runtimeException;
	}
}
```
you miss stacktraces of other exceptions 
```
java.lang.RuntimeException: hello
	at excceptionhandling.Test.main(Test.java:10)
	Suppressed: java.lang.IllegalArgumentException: argument
		at excceptionhandling.Test.main(Test.java:11)
Caused by: java.lang.RuntimeException: reason
	at excceptionhandling.Test.main(Test.java:9)
```

SougandhS:
>compare it to: 

Now its same as what `StringWriter` `PrintWriter` returns

image

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JavaDetailFormattersManager.java
```diff
@@ -390,6 +391,41 @@ private String getDetailFormatterSuperClass(IJavaClassType type) throws DebugExc
...
+					} else {
+						stackTrace.append(this.getClass().getName() +\" \\n\");
+					}
+					for (StackTraceElement element : this.getStackTrace()) {
+						stackTrace.append("\t");
+						stackTrace.append("at "+element.toString()+\" \\n\");
+					}
+					if(this.getSuppressed().length>0) {
+						stackTrace.append("\t");
+						stackTrace.append("Suppressed: ");
```
jukzi, line 408: this impl will not print recursive suppressed/cause throwables. Why don't you just use the printStackTrace()?

jukzi: thats not a bug but a feature. The "missing" lines are only those who are already printed in the parent exception. We see and happy live with that all time. for example 
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3317#issue-2662247656

SougandhS: Cool! then I'll do `printStackTrace()` then

jukzi (approved):

SougandhS:
Will add this in N & N 👍

SougandhS:
Thank you so much @jukzi 😊

PR #613, opened 2025-02-04
Try to fix & clean up VirtualThreadsDebugViewTests

stephan-herrmann:
+ add java23 to build.properties to fix failures on the build server
+ change explicit (duplicate) creation to the normal style 'assert...'
+ allow running it on 23+

### DIFF
--- a/org.eclipse.jdt.debug.tests/build.properties
+++ b/org.eclipse.jdt.debug.tests/build.properties
@@ -26,6 +26,7 @@ bin.includes = plugin.xml,\
                java8/,\
                java9/,\
                java16_/,\
+               java23/,\
                java24/
 source.javadebugtests.jar = test plugin/,\
                             tests/,\
--- a/org.eclipse.jdt.debug.tests/test plugin/org/eclipse/jdt/debug/testplugin/JavaProjectHelper.java
+++ b/org.eclipse.jdt.debug.tests/test plugin/org/eclipse/jdt/debug/testplugin/JavaProjectHelper.java
@@ -190,6 +190,15 @@ public class JavaProjectHelper {
 	}
 
 	/**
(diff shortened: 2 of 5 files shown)


stephan-herrmann:
@SougandhS fyi: although your recent change let the test use Java 23, the test ran only on 24 (see `AutomatedSuite`) and so only in BETA_JAVA24 tests  we detected this failure:
```
junit.framework.AssertionFailedError: Failed to create the 'Two_Three' test project.
'jp' is  'null'
Stack tace:
java.lang.NullPointerException: Cannot invoke "java.net.URL.getFile()" because "localURL" is null
	at org.eclipse.jdt.debug.testplugin.JavaTestPlugin.getFileInPlugin(JavaTestPlugin.java:94)
	at org.eclipse.jdt.debug.tests.AbstractDebugTest.createProject(AbstractDebugTest.java:1039)
	at org.eclipse.jdt.debug.tests.ui.VirtualThreadsDebugViewTests.create23Project(VirtualThreadsDebugViewTests.java:97)
```
Should be fixed by my change in build.properties.

Also duplicate creation of the java project was not optimal :)

SougandhS:
>Should be fixed by my change in build.properties.

Many thanks on this

>Also duplicate creation of the java project was not optimal :)

Noted Thanks 👍

stephan-herrmann:
@SougandhS do you plan any further changes in this area in master? If so, then we should probably apply this fix in master first. OTOH, BETA_JAVA24 is where the fix is currently _needed_.

SougandhS:
> @SougandhS do you plan any further changes in this area in master? If so, then we should probably apply this fix in master first. OTOH, BETA_JAVA24 is where the fix is currently _needed_.

No plans for further changes in master. 👍  you can go ahead with this changes

PR #635, opened 2025-03-07
More locations need to handle version "24"

stephan-herrmann:
Update:
+ StandardVMType
+ unify code from EECompilationParticipant and JavaRuntime

Test currency vis-a-vis JavaCore.latestSupportedJavaVersion()

Also update stale test still mentioning J2SE-1.4

### DIFF
--- a/org.eclipse.jdt.debug.tests/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.debug.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jdt.debug.tests; singleton:=true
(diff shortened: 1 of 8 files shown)


stephan-herrmann:
Observations:
* Needs update after Java 24 GA: org.eclipse.jdt.internal.launching.StandardVMType.getDefaultJavadocLocation(String)
* EECompilationParticipant and JavaRuntime had similar code for converting a VM version to a compliance value
   * EECompilationParticipant.getCompilerCompliance()
       * simple
       * testable (public static)
   * org.eclipse.jdt.launching.JavaRuntime.updateCompliance()
       * more sophisticated (recognizes dot-separated versions)
       * not testable (block hidden in a private method)

I unified that code in EECompilationParticipant and JavaRuntime, by splicing the better detecting from the latter into the former, and call that also from JavaRuntime.

And yes: whenever JavaCore.latestSupportedJavaVersion() is bumped up, 2 new tests will require the necessary updates in this area.

PR #639, opened 2025-03-13
Allow JavaVariableLabelProvider's presentation to be overridden

kysmith-csg:
This mostly reverts a4eae7ff2f3dc3f51234053cc4a9374c2f6f8b51 (#580) for this file, as well as some code cleanup. Instead of making the fLabelProvider final, we initialize it to a default value as before but provide getters/setters so the field can still be private. Details in #580 for why I need this change, but in short it's because our product has a debugger that is tightly coupled to the JDT debugger, basically being an extension of it. Since this probably only affects me, your attention is greatly appreciated.

I don't understand this code enough to know why fLabelProvider is static, given that label providers are not meant to be shared, so I left it alone. Otherwise it would make more sense to have it as an instance field, maybe defined in an overloaded constructor.

I also thought to update the version of this plugin to 3.15 but Eclipse gave me an error when I did that about "no new APIs added" which I don't think is correct, but please let me know.


kysmith-csg:
@iloveeclipse done, I hope this is what you meant

iloveeclipse (approved):

PR #630, opened 2025-03-05
Close HotCodeReplace dialog automatically when process ends

subyssurendran666:
## What it does
Implemented DebugEventListener to monitor debugging events and ensured the popup automatically closes when the thread has successfully terminated.

This PR closes: #596 

## Author checklist

- [ ] I have thoroughly tested my changes
- [ ] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [ ] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #646, opened 2025-03-19
Remove Debug listener on HCR popup close

SougandhS:
Fixes : https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/630/files#r2003358448

Remove debug event listener when HCR failure dialog is closed after thread termination

## Author checklist

### DIFF
--- a/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/HotCodeReplaceErrorDialog.java
+++ b/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/HotCodeReplaceErrorDialog.java
@@ -33,8 +33,8 @@ import org.eclipse.swt.events.TraverseListener;
 import org.eclipse.swt.layout.GridData;
 import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Composite;
-import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Shell;
+import org.eclipse.ui.PlatformUI;
 
 /**
  * An error dialog reporting a problem with a debug
@@ -159,9 +159,14 @@ public class HotCodeReplaceErrorDialog extends ErrorDialogWithToggle implements 
 	@Override
 	public void handleDebugEvents(DebugEvent[] events) {
 		for (DebugEvent event : events) {
-			if (event.getSource() instanceof JDIThread de) {
-				if (de.isTerminated()) {
-					Display.getDefault().asyncExec(() -> this.close());
+			if (event.getSource() instanceof JDIThread jdiThread) {
+				if(!jdiThread.isTerminated()) {
+					continue;
+				}
+				if (jdiThread.getDebugTarget().equals(target.getDebugTarget())) {
+					DebugPlugin.getDefault().removeDebugEventListener(this);
+					PlatformUI.getWorkbench().getDisplay().asyncExec(this::close);
+					return;
 				}
 			}
 		}


SougandhS:
Cc : @iloveeclipse

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/HotCodeReplaceErrorDialog.java
```diff
@@ -161,7 +161,10 @@ public void handleDebugEvents(DebugEvent[] events) {
 		for (DebugEvent event : events) {
 			if (event.getSource() instanceof JDIThread de) {
 				if (de.isTerminated()) {
-					Display.getDefault().asyncExec(() -> this.close());
+					Display.getDefault().asyncExec(() -> {
+						this.close();
```
iloveeclipse, line 165: Looking on this code I wonder why this dialog assumes that the debug event (one from many we might have in the system with multiple debug targets executed at same time) belongs to the **original** debug target we were opened on.

So there is yet another issue here. To reproduce, start two debug targets, let the dialog open in one target and terminate another. Dialog closes which is not what we expect.

iloveeclipse (changes requested):

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/HotCodeReplaceErrorDialog.java
SougandhS: The issue is reproducible. Added a fix to check debug launch prior closing.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/HotCodeReplaceErrorDialog.java
```diff
@@ -160,8 +160,11 @@ public void run() {
 	public void handleDebugEvents(DebugEvent[] events) {
 		for (DebugEvent event : events) {
 			if (event.getSource() instanceof JDIThread de) {
-				if (de.isTerminated()) {
-					Display.getDefault().asyncExec(() -> this.close());
+				if (de.isTerminated() && de.getLaunch().equals(this.target.getLaunch())) {
```
iloveeclipse, line 163: Are we sure that this.target.getLaunch() and de.getLaunch() are always non-null? I see code in debugger that checks that and I see no javadoc saying the launch must be non-null?
Also I believe we don't need launch, we can directly compare debug targets.
Also listener removal shouldn't be async.

Something like this:
```
	@Override
	public void handleDebugEvents(DebugEvent[] events) {
		for (DebugEvent event : events) {
			if (!(event.getSource() instanceof JDIThread jdiThread)) {
				continue;
			}
			if (!jdiThread.isTerminated()) {
				continue;
			}
			if (target.equals(jdiThread.getDebugTarget())) {
				DebugPlugin.getDefault().removeDebugEventListener(this);
				PlatformUI.getWorkbench().getDisplay().asyncExec(this::close);
				return;
			}
		}
	}
```

iloveeclipse (changes requested):

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/HotCodeReplaceErrorDialog.java
SougandhS: Thanks for this Andrey 👍

SougandhS:
Thank you @iloveeclipse

PR #654, opened 2025-03-30
Discover JVM from more vendors #651

### LINKED ISSUE #651: Discover VM installations from other vendors
[RCPTT project](https://github.com/eclipse-rcptt/org.eclipse.rcptt) has a client using a couple of proprietary JVMs. RCPTT Runner has to be able to select any of installed JVMs (the most suitable for current Application Under Test) unattended (without UI). Current VM discovery mechanism  is hard coded, so proprietary VMs are not discovered.

I could:
- contribute a patch hard coding those proprietary vendors in https://github.com/eclipse-jdt/eclipse.jdt.debug/blob/62dd50617c0704aefc2b79476b08d6307b09a6e1/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java#L188 or
- implement extensible discovery mechanism within DetectVMInstallationsJob or
- handle additional VM discovery separately in RCPTT-specific way, without modifying JDT, by adding them with `org.eclipse.jdt.launching.IVMInstallType.createVMInstall(String)`

What is the preferred way?

/label windows

basilevs:
Fixes #651
Follow-up for #231

## What it does
Only JVMs from following vendors had a chance to be automatically discovered on Windows:

- Eclipse Adoptium
- RedHat
- Java (Oracle)

Now the list is extended.

## How to test
To test
- Install  https://aka.ms/download-jdk/microsoft-jdk-21.0.6-windows-x64.msi with default settings for all users
- ensure the new JVM is detected after JDT restart

## Author checklist

/label Windows

### DIFF
--- a/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
+++ b/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/DetectVMInstallationsJob.java
@@ -185,7 +185,11 @@ public class DetectVMInstallationsJob extends Job {
 	@SuppressWarnings("nls")
 	private void computeWindowsCandidates(Collection<File> rootDirectories) {
 		List<String> progFiles = List.of("ProgramFiles", "ProgramFiles(x86)");
-		List<String> subDirs = List.of("Eclipse Adoptium", "RedHat", "Java");
+		/// Could not find directory name for vendors:
+		/// - Alibaba - instructs to unzip in arbitrary directory
+		/// - IBM - each product carries its own installation of JDK
+		/// - SAP - each product carries its own installation of JDK
+		List<String> subDirs = List.of("Eclipse Adoptium", "RedHat", "Java", "Axiom", "Zulu", "BellSoft", "Microsoft", "Amazon Corretto");
 		rootDirectories.addAll(
 		progFiles.stream()
 			.map(name -> System.getenv(name))


basilevs:
@jukzi  PTAL

jukzi (commented):
I would prefer semicolon (";") separator character for windows.
Also https://docs.oracle.com/en/java/javase/17/docs/specs/man/javac.html
states that the platform path separator character is colon(":") on other platforms.

But since i am not a user of this option i don't care much.
Maybe @mickaelistria has an opinion about this PR?

basilevs:
> I would prefer semicolon (";") separator character for windows.

I'll prepare semicolon variant. [Semicolon is allowed in filenames](https://stackoverflow.com/a/3869598/125562) but might be used less often.
> Also https://docs.oracle.com/en/java/javase/17/docs/specs/man/javac.html states that the platform path separator character is colon(":") on other platforms.

This is not a path, but a single directory name. And while the feature is platform specific, I imagine we could eventually use it for Linux too, so we better use a cross-platform separator (because config.ini building is simpler that way).
> But since i am not a user of this option i don't care much. Maybe @mickaelistria has an opinion about this PR?


Overall I'm happy as long as we choose a cross-platform solution.

basilevs:
> I would be fine hardcoding those names of "populer enough vendors" in the code.

> If we're to add a system property to customize VM detection, I think we should aim for a property that would allow to pass full path instead of just path segments.

Full path does not work for my usecase from  #651, because it includes dynamic version segments. If we rely on full path being provided, we delegate a lot of logic from `DetectVMInstallationsJob` to the launcher, and that requires it to be pretty smart. This logic fits much better to JDT, because:

1. it already has the discovery and
2. every client of JDT would benefit from this logic
3. a full path is already accepted by JDT in a form of JAVA_HOME and JDK_HOME, so this feature is already somewhat covered

> This could be more generally useful, on all OSs.

So far Windows is the only platform where vendor name plays any role. On Linux there is a standard location to put all the VMs in, and on MacOS there is a dedicated discovery tool. 

However, I'm worried, that some vendors might start using custom locations or name of Linux, especially when using a custom installer outside of traditional package system.

mickaelistria:
> Full path does not work for my usecase from https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/651, because it includes dynamic version segments

OK, and what about allowing regexps in the system property then?

I'm afraid a system property for 1 particular segment at a particular location is too much customization power for in the end a very narrow configuration margin. I do prefer hardcoding the usual paths here over adding a very restrained (and possibly frustrating) configuration strategy; listing the most probable locations is really what this class is about so this doesn't seem like a hack.

So my preference would be for any of the extreme: either hardcoding, or an entirely powerful setting; everything in between would easily become frustrating IMO.

basilevs:
> OK, and what about allowing regexps in the system property then?

There is no convenient way to apply either Regex or [PathMatcher](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/file/PathMatcher.html)  without scanning the whole directory tree. I wonder if there is a dependency somewhere with Ant-style glob interpreter.

mickaelistria:
Ant-style globs seem to be built-in the JVM API with `FileSystem.getPathMatcher()`

> even Ant patterns are not very good, because even Ant patterns are not very good, because C:\Program Files\**\bin\java.exe will scan quite a lot of software on Windows as all third-party software is installed in C:\Program Files\ will scan quite a lot of software on Windows as all third-party software is installed in C:\Program Files\

Any `**` scans quite a lot, it's usually a bad idea to use it anywhere. But `C:\Program Files\*\bin\java.exe` (with a single star) would be fine to use.

basilevs:
> Ant-style globs seem to be built-in the JVM API with `FileSystem.getPathMatcher()`

No, that api can verify a path, but not path prefix, resulting in non-truncated scans.

mickaelistria:
We can probably do a basic parsing here: split on `/`, and go recursively into children that match the current segment.

jukzi:
lets not overengineer a simple and niche usecase

mickaelistria:
> lets not overengineer a simple and niche usecase

It's not so niche. It won't be long until someone tells "I usually put my all my VMs in folder XXX, how could those be autodetected" (it's more or less what this request is already about), and this can happen with very different locations and also with different OSs, hence why I think we should avoid half-baked solutions right now.
But again, I'd be fine with just hardcoding more paths just as much as with a "full" setting.

basilevs:
A promising registry location: 

    Computer\HKEY_LOCAL_MACHINE\SOFTWARE\JavaSoft\JDK\17.0.10

basilevs:
@jukzi @mickaelistria I've force pushed a hardcoded solution, but I'm worried that `org.eclipse.jdt.internal.launching.StandardVMType.validateInstallLocation(File)`  might throw on non-standard file protections used by Microsoft.

mickaelistria:
+1 to merge

jukzi (approved):

SougandhS (approved):

SougandhS:
@basilevs thanks for the contribution

PR #659, opened 2025-04-03
Change the priority of the categorization

### LINKED ISSUE #658: Lambda code is not properly recognized as project code in debugger
Follow up on https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/583.

Try to debug `org.eclipse.swt.tests.manual.Bug569752_DetectNonDisposedOsResources` with SWT at https://github.com/eclipse-platform/eclipse.platform.swt/pull/1702.

The stack looks like below:

Image

... and with the recent changes to the frame icons there is an unexpected behavior: the two highlighted frames should ideally have different icons, opposite to the current state.

This frame: `0x0000000800c87580.handleEvent(Event) line: not available` is clearly synthetic one, so it shouldn't be shown as "project code", it has not even a line info.

The frame above `Bug569752_DetectNonDisposedOsResources.lambda$0(Display, Button, Event) line: 47`	is, in my understanding, is decorated as a library/non-project code - which is wrong because we clearly see that it matches the current line in our code in which debugger is stopped.

IMO the decoration of both frames should be changed.

gzsombor:
* if marked with custom
* if we have the source file - a test/prod code
* platform code
* synthetic methods
* library code
* unknown

## What it does
Fixes #658

## How to test

Debug the following code:
```
public class Bug {
		var in = new Inner(m -> {
			System.out.println("hello " + m); /// breakpoint here
		});
		for (int i = 0; i < 3; i++) {
			in.innerCall("world " + i);
		}
	}

	record Inner(Consumer<String> li) {
		void innerCall(String msg) {
			li.accept(msg);
		}
	}
}
```
Check that the stack frames are sensible and correctly displayed

## Author checklist

### DIFF
--- a/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/StackFrameCategorizer.java
+++ b/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/StackFrameCategorizer.java
@@ -165,17 +165,24 @@ public class StackFrameCategorizer implements IPreferenceChangeListener {
 			if (isEnabled(CATEGORY_CUSTOM_FILTERED) && custom.match(refTypeName)) {
 				return CATEGORY_CUSTOM_FILTERED;
 			}
+			Category category = categorizeSourceElement(frame);
+			// if the category is prod or test, that's the most relevant.
+			if (category == CATEGORY_PRODUCTION || category == CATEGORY_TEST) {
+				return category;
+			}
+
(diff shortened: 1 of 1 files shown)


iloveeclipse (approved):

iloveeclipse:
@gzsombor : could you add tests for the stack frames categorization to cover known expectations? The change looks fine for this concrete use case, but honestly speaking, it might cause regressions in other use cases I didn't tested manually, and without tests we would only know about regression after someone reports that in the released version.

gzsombor:
There are tests in #81 and I had to modify it due this change. It's not much but touches enough relevant code IMHO.

Issue #661, opened 2025-04-08
Dynamically detect newer JavaSE-N execution environments from Installed JREs


mickaelistria:
Currently, allowing to target JavaSE-N requires that JDT ships support for Java N. However, in some cases, it might not be strictly necessary to get compiler for N in order to target code using library N. The compiler version mostly matters for the source/target release, not so much for the execution environment.
It looks like ExecutionEnvironmentAnalyzer hardcodes the list of possible Java versions. But instead of that, we should consider make it more dynamic and list all the EEs that are known by the IDE, basically all the EEs that are known by the Installed JREs. So for example, if I add a Java installation that tells it's Java 25 (`java -version` returns 25), JDT could automatically list JavaSE-25 as a known Execution Environment; without requiring change in the code.

akurtakov:
This is actually the major point of what was discussed in https://github.com/eclipse-simrel/.github/blob/main/wiki/Planning_Council/2025-03-05.md#java-support-schedule

PR #663, opened 2025-04-10
Some dynamic discovery for environment versions

mickaelistria:
## What it does

Provides some capability for environments that are not yet statically described in JDT.

This is a step towards
https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/661

## How to test

Just ensure no regressions.
Because further guards are in place (eg checking Compiler support for given EE compliance), this would usually be a no-op as for the EE to be returned by the EEManager, the compiler must also declare support for next version.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/environments/EnvironmentsManager.java
+++ b/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/environments/EnvironmentsManager.java
@@ -34,9 +34,9 @@ import org.eclipse.core.runtime.IExtensionPoint;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.preferences.InstanceScope;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences.IPreferenceChangeListener;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences.PreferenceChangeEvent;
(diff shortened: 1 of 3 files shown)


org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/environments/EnvironmentsManager.java
```diff
@@ -253,6 +253,16 @@ private String getExecutionEnvironmentCompliance(IExecutionEnvironment execution
 			return JavaCore.VERSION_1_1;
 		} else if (desc.indexOf("1.0") != -1) { //$NON-NLS-1$
 			return "1.0"; //$NON-NLS-1$
+		} else {
+			// an optimistic strategy for not-yet-declared versions
```
iloveeclipse, line 257: That produces `1` from `25.0.1`, and that is really **guessing**...
I'm not sure this is the right way to guess the version out of EE.
The entire method looks like a big hack.
Why can't we add an extra "JavaCompliance" attribute to `org.eclipse.jdt.launching.executionEnvironments` extension point and specify there all what we need?

mickaelistria: > That produces 1 from 25.0.1, and that is really guessing..

The current code would also return 24 for 1.0.24, so it's not so much worse...

> Why can't we add an extra "JavaCompliance" attribute to org.eclipse.jdt.launching.executionEnvironments extension point and specify there all what we need?

We sure can, but usually, this happens on a branch that takes ages before it is merged...

Issue #674, opened 2025-04-19
[25] Handle version 25 in all standard places


stephan-herrmann:
see previous:
* #551
* #555 
* #598 
* #616 

And let's try to do it all at once :)

PR #675, opened 2025-04-19
[25] Handle version 25 in all standard places #674

stephan-herrmann:
+ bump version of j.launching by +50
+ recognize JavaSE-25 EE and JavaCore.VERSION_25
+ create extension for EE JavaSE-25
+ allow PR build to access the Y-Builds repository
* update build to
  * run on jdk 25
  * create & enable profile test-on-javase-25
* update launcher for local testing: use 24 and add agent jar
* move tests for instance main methods to new project java25

Fixes https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/674


stephan-herrmann:
Weird: I consistently see failure in `ConditionalBreakpointsWithFileClass.testFileConditionalBreakpointforFalse()` (on jenkins, locally on BETA_JAVA25 as well as on master).  Apparently this is caused by `BinaryBasedSourceGenerator` generating a stub class containing "implements java/io/Serializable", i.e., source with a wrong package separator.

At this point I wonder how this test can pass on jenkins for master. Should we actually find a source type and use that for creating the evaluation stub?

To unblock work in the branch I fixed the one location in BinaryBasedSourceGenerator that did not call `.replace('/', '.')` - actually extracted signature conversion to a private method for consistent use.

stephan-herrmann:
> At this point I wonder how this test can pass on jenkins for master. Should we actually find a source type and use that for creating the evaluation stub?

Here's a bold guess: for finding a source type we need a stratum added by the advanced source lookup agent (javaagent-shaded.jar). That guy, however, seems to be unable to weave into the premordial class `File`. Hence a conditional breakpoint in class `java.io.File` needs the binary variant of stub generation. Which brings us back to BinaryBasedSourceGenerator creating illegal syntax.

Next theory: had ecj been more robust in face of the syntax error, creating a useful evaluation stub anyway?

Issue #672, opened 2025-04-18
Deadlock in ExecutionEnvironmentsPreferencePage


basilevs:
# Test steps

- Open Java launch configuration
  - Select an execution environment with radio and combo
  - Click a button "Environments..."
  - In the opened dialog 
    - In Java/Installed JREs/Execution Environments
      - Select another JRE for the environment selected above
      - Click Apply and Close button

# Expected result
No deadlock

# Actual result
```
"main" #1 [259] prio=6 os_prio=31 cpu=320208.67ms elapsed=4971.51s tid=0x000000014f80e200 nid=259 waiting for monitor entry  [0x000000016b885000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at org.eclipse.jdt.internal.launching.environments.EnvironmentsManager.getSupportedExecutionEnvironments(EnvironmentsManager.java:169)
	- waiting to lock <0x000000060141d968> (a org.eclipse.jdt.internal.launching.environments.EnvironmentsManager)
	at org.eclipse.jdt.internal.debug.ui.jres.JREsComboBlock.fillWithWorkspaceProfiles(JREsComboBlock.java:466)
	at org.eclipse.jdt.internal.debug.ui.jres.JREsComboBlock.showPrefPage(JREsComboBlock.java:303)
	at org.eclipse.jdt.internal.debug.ui.jres.JREsComboBlock$3.handleEvent(JREsComboBlock.java:225)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4660)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1657)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1680)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1665)
	at org.eclipse.swt.widgets.Widget.notifyListeners(Widget.java:1394)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:4427)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4003)
	at org.eclipse.jface.window.Window.runEventLoop(Window.java:823)
	at org.eclipse.jface.window.Window.open(Window.java:799)
[...]

PR #681, opened 2025-04-24
Don't use Display.syncExec() for refresh in RuntimeClasspathViewer

iloveeclipse:
The listener may be called from a background thread that already holds some locks. To refresh RuntimeClasspathViewer UI in the current case, syncExec is not needed, asyncExec is enough and it would not acquire the UI lock from an unknown thread.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/672

### DIFF
--- a/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/classpath/RuntimeClasspathViewer.java
+++ b/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/classpath/RuntimeClasspathViewer.java
@@ -64,17 +64,7 @@ public class RuntimeClasspathViewer implements IClasspathViewer {
 
 		@Override
 		public void preferenceChange(PreferenceChangeEvent event) {
-			if (DebugUIPlugin.getStandardDisplay().getThread().equals(Thread.currentThread())) {
-				refresh(true);
-			} else {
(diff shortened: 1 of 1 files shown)


basilevs:
I've seen this pattern of conditional asyncExec before and I don't understand it most of the time. Why do we have to execute the target immeditely if we are in UI thread? Can't we just always schedule the execution for later with `asyncExec()`?

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/classpath/RuntimeClasspathViewer.java
```diff
@@ -64,17 +65,12 @@ public class RuntimeClasspathViewer implements IClasspathViewer {
 
 		@Override
 		public void preferenceChange(PreferenceChangeEvent event) {
-			if (DebugUIPlugin.getStandardDisplay().getThread().equals(Thread.currentThread())) {
+			Display display = DebugUIPlugin.getStandardDisplay();
+			if (display.getThread().equals(Thread.currentThread())) {
```
basilevs, line 69: This is no longer necessary, unconditional async would work.

iloveeclipse:
Thanks @basilevs, I've updated patch.

PR #690, opened 2025-05-01
Force qualifier update to fix build failure

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3006


merks:
@iloveeclipse 

This is needed to fix the I-build:
```
1.  eclipse.jdt.debug/org.eclipse.jdt.debug.ui/pom.xml
   no-classifier: different
      org/eclipse/jdt/internal/debug/ui/launcher/JreResolutionGenerator$ConfigureSeverityResolution.class: different
    The main artifact has been replaced with the baseline version.
    The following attached artifacts have been replaced with the baseline version: [sources]
```

akurtakov:
Looking at the diff 
image
 So this is not compiler caused problem. Merging.

PR #694, opened 2025-05-06
Unable to set a conditional breakpoint in a method with a parameter using generic extends X

### LINKED ISSUE #693: Unable to set a conditional breakpoint in a method with a parameter using generic extends X
Unable to set a conditional breakpoint in a method with a parameter using generic extends X :

Image

The reproduction code :

```
package test;

public class ConditionnalBreakpointWithGeneric<U extends Number> {

    void test(ConditionnalBreakpointWithGeneric<U> conditionnalBreakpointWithGeneric) {
        System.out.println("T");
    }

        ConditionnalBreakpointWithGeneric<Integer> conditionnalBreakpointWithGeneric = new ConditionnalBreakpointWithGeneric<Integer>();
        conditionnalBreakpointWithGeneric.test(conditionnalBreakpointWithGeneric);
        System.out.println("");
    }
}
```

This is due to commit e782ff0b726016b5a8132e0479562d7f2007f151 using QN_OBJECT

RoiSoleil:
Fix #693


RoiSoleil:
@iloveeclipse @gayanper

iloveeclipse:
@RoiSoleil : the test fails:

```
org.eclipse.jdt.debug.tests.eval.LambdaVariableTest.testEvaluate_Bug567801_VariableWithIntersectionTypeArgument_MustEvaluationWithCorrectType
The evaluation of '" + snippet + "'  should not have errors : [T cannot be resolved to a type, This lambda expression refers to the missing type T]
```

RoiSoleil:
To fix this test fail i need to specify the generic method type in RemoteEvaluatorBuilder#build but there is no API on IEvaluationContext to do so.  Could you take a look because i'm not sure i can handle it ...

org.eclipse.jdt.debug/eval/org/eclipse/jdt/internal/debug/eval/ast/engine/ASTEvaluationEngine.java
```diff
@@ -392,7 +392,11 @@ private void scanAndFixSignature(String genericSignature, String erasureSignatur
 		 * and also generic signature such as Ljava/util/function/Predicate<+Ljava/util/List<Ljava/lang/Integer;>;>; Ljava/util/Comparator<-TT;>;
 		 * which will fail to properly resolved to the type.
 		 */
-		if (genericSignature.startsWith(String.valueOf(Signature.C_TYPE_VARIABLE)) || genericSignature.startsWith(String.valueOf(Signature.C_CAPTURE))
+		if (genericSignature.startsWith(String.valueOf(Signature.C_TYPE_VARIABLE))) {
+			fixedSignature.append(genericSignature.substring(String.valueOf(Signature.C_TYPE_VARIABLE).length(), genericSignature.length() - 1));
```
gayanper, line 396: could you provide the value you get in the generic signature with your test case which you provided this fix for ?

gayanper:
> To fix this test fail i need to specify the generic method type in RemoteEvaluatorBuilder#build but there is no API on IEvaluationContext to do so. Could you take a look because i'm not sure i can handle it ...

The idea of the original fix was to avoid using generic types like in the failing test scenario since we are unable to resolve them in debugger context. The reason is we are unable to parse the snippet in such a way where we can resolve generics.

So now with this fix, it seems the snippet might be ending up with a generic type which was not replaced, if you can debug the final snippet after applying the `scanAndFixSignature` invocation, We can find out what is happening.

gayanper:
@RoiSoleil thanks for coming up with a PR, it is really appreciated the time you put into coming up with this fix.

PR #716, opened 2025-06-10
[25] update version to ensure greater than master

stephan-herrmann:
Master recently bumped o.e.j.debug to 3.23.100. Now we go to 3.23.150 in beta branch to ensure that this is understood as being newer than master (we do have changes in the branch since dc73593d91bc0d89f7cef76878891bf245e61850 ).

PR #717, opened 2025-06-10
Merge master to BETA_JAVA25

stephan-herrmann:
#716 failed because o.e.j.debug.ui has a higher version in master than in BETA_JAVA25.

Hence merging latest from master into the beta branch.


stephan-herrmann:
@jarthana may I bother you to check this PR?

Something seems to have gone wrong, as this merge only shows one parent :(

During previous such merges I directly pushed the local merge to origin/BETA_JAVAXY, which worked fine.

Then I saw that you found a way to avoid directly pushing, so I tried the same. Being unsure what github would actually do I chose squash-and-merge, which apparently squashed the merge commit into a regular commit.

Did you previously chose rebase-and-merge? I didn't dare to choose that, since I figured that a merge commit can impossibly be rebased!

After all that locally merging master again into BETA_JAVA25 worked as an empty merge commit, so git understands that those changes are already in.

Perhaps we should add something like that as another PR, to set a new baseline between branches?

Unfortunately even with the additional merge commit changes from between your last such merge and now (4 commits, incl. 1 version bump) will have blame assignment pointing to my squashed merge :(

If really needed, we would need help from admins so we can re-create the branch from before my merge (with temp rename to avoid data loss). Do we need to?

stephan-herrmann:
> I chose squash-and-merge, which apparently squashed the merge commit into a regular commit.

I should have known better, since I already successfully applied this to eliminate unwanted merge commits from regular PRs.

Sorry for the mess!

jarthana:
I am typing from my mobile. So, I don't have a clear view of things. When I merge, I enable the merge option from the repo setting page, which I believe is accessible only to the project leads. Perhaps that's the problem here. I can have a closer look tomorrow and if that's the issue, we can reset the beta branch and try again.

stephan-herrmann:
> I am typing from my mobile. So, I don't have a clear view of things. 

Thanks for answering anyway :)

> When I merge, I enable the merge option from the repo setting page, which I believe is accessible only to the project leads. 

That's interesting. Well I can see some repo settings, but nothing that obviously applies to this use case.

> Perhaps that's the problem here. I can have a closer look tomorrow and if that's the issue, we can reset the beta branch and try again.

That'd be great.
Sorry, again, for the mess. Should've kept the previously successful approach.

stephan-herrmann:
For posterity: this bogusly squashed change has been removed from the history. Next attempt is #718

PR #718, opened 2025-06-11
Merge master into BETA_JAVA25

stephan-herrmann:
Bring in recent version bumps to prepare for a P-build.


stephan-herrmann:
@jarthana my goal is to resolve all version issues in the branch before the next Y-build, which I believe is scheduled for Thu 2:00 pm UTC. So, assuming test success, it's fully sufficient if you merge this PR Thu morning your time, which still gives me enough time add add branch specific version bumps where needed (which then doesn't require your help).
thanks

Issue #721, opened 2025-06-12
CI-build time exceeds the timeout


stephan-herrmann:
Since recently many build jobs for jdt.debug are being aborted at the timeout of 40 minutes, while previously a typical time to completion was below 30 minutes.

The list on https://ci.eclipse.org/jdt/job/eclipse.jdt.debug-github/job/master/ indicates that build times went up after https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/692 - already the PR build was aborted - but I have no idea if that change possibly brought such performance regression or whether all this is based only on "bad weather" on our build infra.

Anyone have a guess?

@SougandhS @subyssurendran666  @HannesWell @jarthana

stephan-herrmann:
Should we increase the timeout on jenkins just for now?

SougandhS:
> Should we increase the timeout on jenkins just for now?

Maybe infra issues. But we can try by increasing timeout and lets see.

iloveeclipse:
Most likely this is a side effect of https://github.com/eclipse-platform/eclipse.platform.ui/pull/3025 & waitForJobs() waiting for reconciler jobs now.

PR #722, opened 2025-06-12
Avoid waiting for reconciler job

iloveeclipse:
Not so nice workaround for
https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/721.

After https://github.com/eclipse-platform/eclipse.platform.ui/pull/3025 every opened editor means there will be a reconciler job running, which is not what we do NOT want to wait for. AbstractReconciler.class is the family object we can check for, so add that family to the list of excluded jobs we don't care.


SougandhS:
No more jenkins failure 👍

Issue #695, opened 2025-05-07
[25] Fix and re-enable ConditionalBreakpointsWithFileClass.testFileConditionalBreakpointforTrue()


stephan-herrmann:
To unblock #675 I had to disable one test, that failed on jenkins while passing locally.

See https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/675#issuecomment-2860000482 f. 

To add to the weirdness, initially we saw testFileConditionalBreakpointforFalse failing, for which I added a fix in BinaryBasedSourceGenerator.  Then testFileConditionalBreakpointforTrue failed consistently on jenkins.

PR #729, opened 2025-06-21
[25] Fix and re-enable ConditionalBreakpointsWithFileClass.testFileConditionalBreakpointforTrue()

stephan-herrmann:
+ try re-enabling (couldn't reproduce locally, so no fix)

Fixes https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/695

### DIFF
--- a/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/breakpoints/ConditionalBreakpointsWithFileClass.java
+++ b/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/breakpoints/ConditionalBreakpointsWithFileClass.java
@@ -51,8 +51,7 @@ public class ConditionalBreakpointsWithFileClass extends AbstractDebugTest {
 		}
 	}
 
-	// FIXME disabled to unblock BETA_JAVA25
-	public void _testFileConditionalBreakpointforTrue() throws Exception {
+	public void testFileConditionalBreakpointforTrue() throws Exception {
 		String typeName = "FileConditionSnippet2";
 		IJavaLineBreakpoint bp3 = createLineBreakpoint(20, typeName);
 		IJavaThread mainThread = null;


stephan-herrmann:
Still fails on jenkins with "Thread should be suspended".

Debugging this locally, I was surprised that I could observe multiple worker threads ("JDI Expression Evaluation Event Dispatch") calling `JavaBreakpoint.handleEvent()`. This begs the question if some racing could occur between those threads, but I haven't investigated this in depth. Might have been an artifact of suspending on my own breakpoints in jdt.debug.

stephan-herrmann:
> In fact running the test on my local machine (from the IDE) it _consistently passes_.

For good measure I also ran a local maven build, which showed 6 failures and 2 errors, but not the one we are looking for: both tests in ConditionalBreakpointsWithFileClass passed.

SougandhS:
>Are you arguing for selective re-enabling (only below 25)?
>Still I wouldn't like the notion of "permanent", because that would mean we acknowledge loss of functionality.

My suggestion was to disable the test in `jdt.debug.tests.AutomatedSuite` for 25

SougandhS:
>My suggestion was to disable the test in jdt.debug.tests.AutomatedSuite for 25

eg
```
		if (!JavaProjectHelper.isJava25_Compatible()) { // Consistent failure on Java 25
			addTest(new TestSuite(ConditionalBreakpointsWithFileClass.class));
		}
```

stephan-herrmann:
> > Are you arguing for selective re-enabling (only below 25)?
> > Still I wouldn't like the notion of "permanent", because that would mean we acknowledge loss of functionality.
> 
> My suggestion was to disable the test in `jdt.debug.tests.AutomatedSuite` for 25

The test *is* currently disabled. That's why I asked if your goal is to enable this test at least for 24 and below. Anyway, let's continue in 4.38 then.

SougandhS:
I have merged https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/770 
We will have a clean java 25 results in next night build

Issue #744, opened 2025-07-28
Add JavaSE-25 EE to launching


trancexpress:
See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4259, we want to test with OpenJDK 25 early access, we run into JRE .classpath container problems when setting `JavaSE-25`.

See e.g.: https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/426

laeubi:
I see same errors as reported here:
- https://github.com/eclipse-pde/eclipse.pde/issues/1227

so starting an ibuild to see if it helps:
- https://ci.eclipse.org/releng/job/Builds/job/I-build-4.37/87/

PR #746, opened 2025-07-28
Allow selecting JavaSE-25 from the environments list for debugging

iloveeclipse:
Fixes https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/744


trancexpress:
Sorry, I must have missed something from the previous PR.

iloveeclipse:
> Sorry, I must have missed something from the previous PR.

Not only you, Alex too in his original change :-)

akurtakov:
The process of adding new Java version is very convoluted and involves way too many places . My plan was to removing the need to edit manually in leaves (like in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1966 ) and use experience gathered from doing so for trying to automate more in other places (e.g. BREEs) . The stars laughed at my plans and decided smth totally different for me now so that battle would be for someone else.

iloveeclipse:
With https://download.eclipse.org/eclipse/downloads/drops4/I20250728-1800 the build is successful (except warnings below).
Reported "new" warnings are neither new nor related:
- JavaConnectTab extends non-API type AbstractJavaMainTab
- The API problem filter for: 'RuntimeClasspathViewer illegally extends TableViewer' is no longer used

Meging.

PR #750, opened 2025-07-29
Make determination of execution environment compliance generic

laeubi:
Currently one needs to add a new if for each new java version while the value is only matched against the number in the string.

This now uses a pattern and check the result is larger than Java 1.8 version and then returns the number as a string so it automatically adapt to new java versions.


stephan-herrmann:
From all that I can, this will only facilitate future maintenance, it doesn't add any new capability, right?
Note that this particular method is exclusively used for _sorting_ EE entries.

Issue #760, opened 2025-08-08
Selection of "best matching" JRE is inconsitent (if not wrong)


laeubi:
This is a variant of

- https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/756

and discovered here

- https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/753#issuecomment-3163344357

As described in [the issue](https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/756):

> even without any multi-release configuration there is no 100% guarantee that a higher JVM matches the exact behavior of the selected version; think of any modules / classes / methods that are effectively _removed_ from a JVM at a _higher version_, whereas the selected lower version would still have it

This implies that when selecting a JRE for a given EE, the risk that something is removed or changes is _higher_ the larger the gap is between the desired runtime and a perfect match.

But currently `org.eclipse.jdt.internal.launching.JREContainerInitializer.resolveVM(IExecutionEnvironment)` has some code that simply selects the **default JVM if it matches** regardless of other choices. If that one do not match it even select the first (what is random as it only depends on in what order installs are added). So assume we want a `EE11` and have a default JVM of `Java 24` and another one of `Java 12` the java 24 is chosen that at has a much higher chance of removal that `Java 12`.

Also this is not reflected elsewhere, e.g. `org.eclipse.jdt.internal.debug.ui.jres.JREsEnvironmentComparator` only sort by perfect match and then _by name_ of the JRE giving a false impression of ordering in the UI. The we have `org.eclipse.jdt.internal.ui.refactoring.reorg.PasteAction.TextPaster.computeLatestVM()` that already uses an ordering of jvms to find the best matching (that is the first with lowest).

Because of this I think we should have a new method as an alternative to `EE.getCompatibleVMs()` that is `EE.getCompatibleVM()` and adhere to the rule of returning the lowest compatible JVM, then also `JREsEnvironmentComparator` should be adjusted to reflect this ordering as well, so that the most likely selected one is first regardless of name.

stephan-herrmann:
Has this been resolved by #761?

Issue #719, opened 2025-06-12
[Test] org.eclipse.jdt.debug.tests.breakpoints.ThreadNameChangeTests failed


SougandhS:
IBuild : I20250611-0310

https://download.eclipse.org/eclipse/downloads/drops4/I20250611-0310/testresults/html/org.eclipse.jdt.debug.tests_ep437I-unit-linux-x86_64-java24_linux.gtk.x86_64_24.html

org.eclipse.jdt.debug.tests.breakpoints.ThreadNameChangeTests

```
unexpected number of events: [] expected:<1> but was:<0>

junit.framework.AssertionFailedError: unexpected number of events: [] expected:<1> but was:<0>
  at junit.framework.Assert.fail(Assert.java:57)
  at junit.framework.Assert.failNotEquals(Assert.java:329)
  at junit.framework.Assert.assertEquals(Assert.java:78)
  at junit.framework.Assert.assertEquals(Assert.java:234)
  at junit.framework.TestCase.assertEquals(TestCase.java:377)
  at org.eclipse.jdt.debug.tests.breakpoints.ThreadNameChangeTests.testListenToThreadNameChange(ThreadNameChangeTests.java:88)
  at java.base/java.lang.reflect.Method.invoke(Method.java:565)
  at junit.framework.TestCase.runTest(TestCase.java:177)
  at junit.framework.TestCase.runBare(TestCase.java:142)
  at org.eclipse.jdt.debug.tests.AbstractDebugTest.runBare(AbstractDebugTest.java:2818)
  at junit.framework.TestResult$1.protect(TestResult.java:122)
  at junit.framework.TestResult.runProtected(TestResult.java:142)
  at junit.framework.TestResult.run(TestResult.java:125)
  at junit.framework.TestCase.run(TestCase.java:130)
  at junit.framework.TestSuite.runTest(TestSuite.java:241)
  at junit.framework.TestSuite.run(TestSuite.java:236)
  at junit.framework.TestSuite.runTest(TestSuite.java:241)
  at org.eclipse.jdt.debug.tests.DebugSuite$1.run(DebugSuite.java:61)
  at java.base/java.lang.Thread.run(Thread.java:1447)
```

SougandhS:
Showed up again : 
[...]

PR #761, opened 2025-08-09
Select the best matching JRE for an EE from the list of compatible JVMs

laeubi:
Currently the selection of a compatible JRE for an EE seems quite random for the user, e.g. it can happen that for an EE10 with some JRE11/17/24 it chooses 24. This is actually a bad choice as newer java might _remove_ API during release, so the _lowest_ possible java version among others would be the best choice here.

This now do the following:

- introduce a new method IExecutionEnvironment#getCompatibleVM that selects the best matching from the set of selected JVMs (but ignoring the default)
- in the JREContainerInitializer check if there is a default vm selected, then use that
- if not ask the EE for a compatible one (with the new method)

See
- https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/760

### DIFF
--- a/org.eclipse.jdt.launching/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.launching/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jdt.launching; singleton:=true
-Bundle-Version: 3.23.500.qualifier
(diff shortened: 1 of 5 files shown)


SougandhS:
Unrelated test failure : https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/719

jjohnstn (approved):
Looks good.  The old code would return the first installed VM if nothing compatible was found but that isn't correct and the code should force the end-user to install an appropriate JRE.

laeubi:
Thanks for looking into this!

PR #765, opened 2025-08-18
Improve performance of AbstractJavaLaunchConfigurationDelegate

### LINKED ISSUE #764: Improve AbstractJavaLaunchConfigurationDelegate.getClasspathAndModulepath/getClasspath
These methods unnecessarily create several lists.
[VS Code Java LS](https://github.com/eclipse-jdtls/eclipse.jdt.ls) often calls these methods - https://github.com/eclipse-jdtls/eclipse.jdt.ls/blob/260ac6ca30a65a88df3b4d587c8988e129ba5840/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/commands/ProjectCommand.java#L435

snjeza:
Fixes #764 

Improve AbstractJavaLaunchConfigurationDelegate.getClasspathAndModulepath/getClasspath

The PR can significantly improve performance when working with large projects. 

cc @rgrunber

### DIFF
--- a/org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/AbstractJavaLaunchConfigurationDelegate.java
+++ b/org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/AbstractJavaLaunchConfigurationDelegate.java
@@ -23,6 +23,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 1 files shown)


rgrunber:
I wouldn't mind having a look. Seems like just using sets is the right approach.

rgrunber (commented):
@snjeza , Is the problem `List.contains(..)` or that there's duplicated large lists ? I assume this code exists to guarantee the order of entries in the Classpath UI. See https://bugs.eclipse.org/bugs/show_bug.cgi?id=521929 (This is where 756cd5421094304609942fa103ca8214165fde6e was introduced) . Wouldn't this break that ? For me it doesn't break, but Set can't guarantee any order.

Screenshot From 2025-08-20 15-58-24

rgrunber (approved):
Yeah I meant whatever was calling `contains(..)` was probably inefficent.

You're right, `LinkedHashSet` should preserve the order things are added (for `toArray`) without having to keep using `ArrayList`.

I would just adjust the commit message :

```
Improve performance of AbstractJavaLaunchConfigurationDelegate

- getClasspathAndModulepath & getClasspath should use a LinkedHashSet to
  efficiently add unique elements while preserving order
```

PR #821, opened 2025-12-04
[26] Adjusting some tests

stephan-herrmann:
+ applets are gone
+ don't enable preview for 25


SougandhS:
Given that Java applets are obsolete, shouldn't we disable all tests related to / use applets for Java 26 ?

PR #831, opened 2025-12-22
Allow stepping through disassembly

danthe1st:
## What it does
This PR allows stepping through the bytecode of disassembled classes in the class file editor.

~~This is just an unfinished prototype but I wanted to create a (draft) PR early to be able to get feedback and show the basic approach.~~

This depends on another PR in jdt.ui (https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2708)

image
image

@SougandhS Since you have worked on similar things recently, I think you might be interested in that.

## How to test

- Start Eclipse with this change as well as [this patch](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2708)
- Step through some code that steps into libraries without a source attachment
- It automatically jumps to the relevant bytecode instruction(s) corresponding to the stack frame
- Stepping through normal (non-library) classes and library classes with source attachments should still work

## Author checklist

### DIFF
--- a/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/sourcelookup/ClassFileEditorHighlightingTest.java
+++ b/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/sourcelookup/ClassFileEditorHighlightingTest.java
@@ -1,0 +1,365 @@
+/*******************************************************************************
+ * Copyright (c) 2026, Daniel Schmid and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Daniel Schmid - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.jdt.debug.tests.sourcelookup;
+
+import java.io.IOException;
+import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Locale;
+
+import javax.tools.DiagnosticCollector;
+import javax.tools.JavaCompiler;
+import javax.tools.JavaCompiler.CompilationTask;
+import javax.tools.JavaFileObject;
+import javax.tools.SimpleJavaFileObject;
+import javax.tools.StandardJavaFileManager;
+import javax.tools.ToolProvider;
+
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IResource;
+import org.eclipse.core.runtime.ILogListener;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.debug.core.ILaunchConfiguration;
+import org.eclipse.debug.core.ILaunchConfigurationWorkingCopy;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.debug.core.IJavaStackFrame;
+import org.eclipse.jdt.debug.core.IJavaThread;
+import org.eclipse.jdt.debug.testplugin.JavaProjectHelper;
+import org.eclipse.jdt.debug.tests.TestUtil;
+import org.eclipse.jdt.debug.tests.ui.AbstractDebugUiTests;
+import org.eclipse.jdt.internal.ui.javaeditor.ClassFileEditor;
+import org.eclipse.jdt.launching.IJavaLaunchConfigurationConstants;
+import org.eclipse.jdt.ui.jarpackager.IJarExportRunnable;
+import org.eclipse.jdt.ui.jarpackager.JarPackageData;
+import org.eclipse.swt.custom.StyleRange;
+import org.eclipse.swt.custom.StyledText;
+import org.eclipse.swt.custom.StyledTextContent;
+import org.eclipse.ui.IWorkbenchWindow;
+import org.eclipse.ui.PlatformUI;
+import org.junit.Assume;
+
+public class ClassFileEditorHighlightingTest extends AbstractDebugUiTests {
+
+	private static final long TIMEOUT = 10_000L;
+
+	private static final String CLASS_NAME = "NoSources";
+	private static final String CLASS_CONTENTS = """
+			public class NoSources {
+				private static int i = 0;
+				public static void main(String[] args) {
+					i++;
+					System.out.println(i);
+				}
+			}
+			""";
+
+	public ClassFileEditorHighlightingTest(String name) {
+		super(name);
+	}
+
+	@Override
+	protected boolean enableUIEventLoopProcessingInWaiter() {
+		return true;
+	}
+
+	@Override
+	public void tearDown() throws Exception {
+		try {
+			closeAllEditors();
+		} finally {
+			super.tearDown();
+		}
+	}
+
+	public void testDisplaySourceWithClassFileEditorHighlightsLine() throws Exception {
+		IJavaProject project = createProjectWithNoSources(CLASS_NAME, CLASS_CONTENTS, true);
+		IJavaThread thread = null;
+		try {
+			ILaunchConfiguration config = createLaunchConfiguration(project, CLASS_NAME);
+			createLineBreakpoint(project.findType(CLASS_NAME), 4);
+			thread = launchToBreakpoint(config);
+
+			stepOver((IJavaStackFrame) thread.getTopStackFrame());
+			expectHighlightedText("     8  getstatic java.lang.System.out : java.io.PrintStream [13]", TIMEOUT);
+
+			stepOver((IJavaStackFrame) thread.getTopStackFrame());
+			expectHighlightedText("    17  return", TIMEOUT);
+
+			resumeAndExit(thread);
+			expectNothingHighlighted(TIMEOUT);
+		} finally {
+			terminateAndRemove(thread);
+			removeAllBreakpoints();
+			project.getProject().delete(true, null);
+		}
+	}
+
+	public void testContinueAfterEditorClosed() throws Exception {
+		IJavaProject project = createProjectWithNoSources(CLASS_NAME, CLASS_CONTENTS, true);
+		IJavaThread thread = null;
+		List<IStatus> errors = Collections.synchronizedList(new ArrayList<>());
+		ILogListener logListener = new ILogListener() {
+			@Override
+			public void logging(IStatus status, String plugin) {
+				if (status.matches(IStatus.ERROR)) {
+					errors.add(status);
+				}
+			}
+		};
+		try {
+			ILaunchConfiguration config = createLaunchConfiguration(project, CLASS_NAME);
+			createLineBreakpoint(project.findType(CLASS_NAME), 4);
+			thread = launchToBreakpoint(config);
+
+			expectHighlightedText("     0  getstatic NoSources.i : int [7]", TIMEOUT);
+
+			closeAllEditors();
+			Platform.addLogListener(logListener);
+			resumeAndExit(thread);
+			assertEquals(List.of(), errors);
+
+		} finally {
+			terminateAndRemove(thread);
+			removeAllBreakpoints();
+			project.getProject().delete(true, null);
+			Platform.removeLogListener(logListener);
+		}
+	}
+
+	public void testConstructorInPackage() throws Exception {
+		IJavaProject project = createProjectWithNoSources("ClassOne", """
+				public class ClassOne {
+
+					public static void main(String[] args) {
+						ClassOne co = new ClassOne();
+						co.method1();
+					}
+
+					public void method1() {
+						method2();
+					}
+
+					public void method2() {
+						method3();
+					}
+
+					public void method3() {
+						System.out.println("ClassOne, method3");
+					}
+				}
+				""");
+		createMethodBreakpoint(project.findType("ClassOne"), "<init>", "()V", true, false);
+		IJavaThread thread = null;
+		try {
+			ILaunchConfiguration config = createLaunchConfiguration(project, "ClassOne");
+			thread = launchToBreakpoint(config);
+			expectHighlightedText("    0  aload_0 [this]", TIMEOUT);
+		} finally {
+			terminateAndRemove(thread);
+			removeAllBreakpoints();
+			project.getProject().delete(true, null);
+		}
+	}
+
+	public void testDisplaySourceWithClassFileEditorHighlightsLineInConstructor() throws Exception {
+		IJavaProject project = createProjectWithNoSources("MethodCall", """
+				public class MethodCall {
+
+					private int i;
+					private int sum = 0;
+
+					public static void main(String[] args) {
+						MethodCall mc = new MethodCall();
+						mc.go();
+					}
+
+					public void go() {
+						calculateSum();
+					}
+
+					protected void calculateSum() {
+						sum += i;
+					}
+				}
+				""");
+		createMethodBreakpoint(project.findType("MethodCall"), "<init>", "()V", true, false);
+		IJavaThread thread = null;
+		try {
+			ILaunchConfiguration config = createLaunchConfiguration(project, "MethodCall");
+			thread = launchToBreakpoint(config);
+			expectHighlightedText("     0  aload_0 [this]", TIMEOUT);
+		} finally {
+			terminateAndRemove(thread);
+			removeAllBreakpoints();
+			project.getProject().delete(true, null);
+		}
+	}
+
+	public void testClassFileWithoutDebuggingInformation() throws Exception {
+		IJavaProject project = createProjectWithNoSources(CLASS_NAME, CLASS_CONTENTS);
+		IJavaThread thread = null;
+		try {
+			ILaunchConfiguration config = createLaunchConfiguration(project, CLASS_NAME);
+			ILaunchConfigurationWorkingCopy workingCopy = config.getWorkingCopy();
+			workingCopy.setAttribute(IJavaLaunchConfigurationConstants.ATTR_STOP_IN_MAIN, true);
+			config = workingCopy.doSave();
+			thread = launchToBreakpoint(config);
+
+			String[] expectedHighlights = """
+					     0  getstatic NoSources.i : int [7]
+					     3  iconst_1
+					     4  iadd
+					     5  putstatic NoSources.i : int [7]
+					     8  getstatic java.lang.System.out : java.io.PrintStream [13]
+					    11  getstatic NoSources.i : int [7]
+					    14  invokevirtual java.io.PrintStream.println(int) : void [19]
+					    17  return
+					""".split(System.lineSeparator());
+
+			for (int i = 0; i < expectedHighlights.length; i++) {
+				expectHighlightedText(expectedHighlights[i], TIMEOUT);
+
+				if (i < expectedHighlights.length - 1) {
+					stepOver((IJavaStackFrame) thread.getTopStackFrame());
+				}
+			}
+		} finally {
+			terminateAndRemove(thread);
+			removeAllBreakpoints();
+			project.getProject().delete(true, null);
+		}
+	}
+
+	private IJavaProject createProjectWithNoSources(String className, String contents) throws Exception {
+		return createProjectWithNoSources(className, contents, false);
+	}
+
+	private IJavaProject createProjectWithNoSources(String className, String contents, boolean generateLineInfo) throws Exception {
+		IJavaProject javaProject = JavaProjectHelper.createJavaProject("ClassFileEditorHighlightingTest", JavaProjectHelper.BIN_DIR);
+		IProject project = javaProject.getProject();
+		project.getFolder(LAUNCHCONFIGURATIONS).create(true, true, null);
+
+		IFile classFile = project.getFile(className + ".class");
+
+		String debug = generateLineInfo ? "lines" : "none";
+		compileWithJavac(className, contents, List.of("-g:" + debug, "-d", project.getLocation().toString(), "--release", "8"));
+		classFile.refreshLocal(IResource.DEPTH_ONE, null);
+
+		IFile lib = project.getFile("lib.jar");
+		jarClassFile(lib, classFile);
+
+		JavaProjectHelper.addLibrary(javaProject, lib.getFullPath());
+		waitForBuild();
+		return javaProject;
+	}
+
+	private static void jarClassFile(IFile jar, IFile classFile) {
+		JarPackageData data = new JarPackageData();
+		data.setJarLocation(jar.getLocation());
+		data.setBuildIfNeeded(true);
+		data.setOverwrite(true);
+		data.setElements(new Object[] { classFile });
+		data.setExportClassFiles(true);
+
+		IStatus status = callInUi(() -> {
+			IWorkbenchWindow window = PlatformUI.getWorkbench().getActiveWorkbenchWindow();
+			IJarExportRunnable op = data.createJarExportRunnable(window.getShell());
+			window.run(false, false, op);
+			return op.getStatus();
+		});
+		if (status.getSeverity() == IStatus.ERROR) {
+			fail("Creating jar failed: " + status.getMessage());
+		}
+	}
+
+	private static void expectHighlightedText(String expectedHighlightedText, long timeout) throws InterruptedException {
+		long s = System.currentTimeMillis();
+		while (System.currentTimeMillis() - s < timeout) {
+			List<String> highlighted = callInUi(ClassFileEditorHighlightingTest::getClassEditorHighlightedText);
+			if (List.of(expectedHighlightedText).equals(highlighted)) {
+				break;
+			}
+			TestUtil.runEventLoop();
+			Thread.sleep(50L);
+		}
+		List<String> highlighted = callInUi(ClassFileEditorHighlightingTest::getClassEditorHighlightedText);
+		assertEquals("Timed out while waiting on highlighting", Arrays.asList(expectedHighlightedText), highlighted);
+	}
+
+	private static void expectNothingHighlighted(long timeout) throws InterruptedException {
+		long s = System.currentTimeMillis();
+		while (System.currentTimeMillis() - s < timeout) {
+			List<String> highlighted = callInUi(ClassFileEditorHighlightingTest::getClassEditorHighlightedText);
+			if (highlighted.isEmpty()) {
+				break;
+			}
+			TestUtil.runEventLoop();
+			Thread.sleep(50L);
+		}
+		List<String> highlighted = callInUi(ClassFileEditorHighlightingTest::getClassEditorHighlightedText);
+		assertEquals("Timed out while waiting on highlighting", List.of(), highlighted);
+	}
+
+	private static List<String> getClassEditorHighlightedText() {
+		List<String> highlighted = new ArrayList<>();
+		StyledText noSourceTextWidget = null;
+		ClassFileEditor editor = (ClassFileEditor) getActivePage().getActiveEditor();
+		if (editor != null) {
+			noSourceTextWidget = editor.getNoSourceTextWidget();
+		}
+		if (noSourceTextWidget != null) {
+			StyleRange[] styleRanges = noSourceTextWidget.getStyleRanges();
+			StyledTextContent content = noSourceTextWidget.getContent();
+			for (StyleRange styleRange : styleRanges) {
+				String highlightedText = content.getTextRange(styleRange.start, styleRange.length);
+				highlighted.add(highlightedText);
+			}
+		}
+		return highlighted;
+	}
+
+	private static void compileWithJavac(String className, String source, List<String> compilerOptions) throws IOException {
+		JavaFileObject fileObject = new SourceJavaFileObject(className, source);
+		JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
+		Assume.assumeNotNull(compiler);
+
+		try (StandardJavaFileManager fileManager = compiler.getStandardFileManager(null, Locale.ROOT, StandardCharsets.UTF_8)) {
+			DiagnosticCollector<JavaFileObject> collector = new DiagnosticCollector<>();
+			CompilationTask task = compiler.getTask(null, fileManager, collector, compilerOptions, null, List.of(fileObject));
+			Boolean result = task.call();
+			assertTrue(String.valueOf(collector.getDiagnostics()), result);
+		}
+	}
+
+	private static class SourceJavaFileObject extends SimpleJavaFileObject {
+
+		private final String code;
+
+		protected SourceJavaFileObject(String name, String code) {
+			super(URI.create("string:///" + name.replace('.', '/') + Kind.SOURCE.extension), Kind.SOURCE);
+			this.code = code;
+		}
+
+		@Override
+		public CharSequence getCharContent(boolean ignoreEncodingErrors) throws IOException {
+			return code;
+		}
+	}
+}
--- a/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JavaStackFrameEditorPresenter.java
+++ b/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JavaStackFrameEditorPresenter.java
@@ -24,6 +24,7 @@ import org.eclipse.jdt.core.dom.LambdaExpression;
 import org.eclipse.jdt.internal.debug.core.JDIDebugPlugin;
 import org.eclipse.jdt.internal.debug.core.model.JDIStackFrame;
 import org.eclipse.jdt.internal.debug.ui.actions.ToggleBreakpointAdapter;
+import org.eclipse.jdt.internal.ui.javaeditor.ClassFileEditor;
 import org.eclipse.jdt.ui.JavaUI;
 import org.eclipse.jface.text.BadLocationException;
 import org.eclipse.jface.text.IDocument;
@@ -39,14 +40,18 @@ import org.eclipse.ui.texteditor.ITextEditor;
 public class JavaStackFrameEditorPresenter implements IDebugEditorPresentation {
 
 	@Override
-	public boolean addAnnotations(IEditorPart editor, IStackFrame frame) {
+	public boolean addAnnotations(IEditorPart editorPart, IStackFrame frame) {
+		if (!(frame instanceof JDIStackFrame jdiFrame && editorPart instanceof ITextEditor textEditor)) {
+			return false;
+		}
+		IDocument document = getDocument(textEditor);
+		if (document == null) {
+			return false;
+		}
 		try {
-			if (editor instanceof ITextEditor textEditor && frame instanceof JDIStackFrame jdiFrame
-					&& org.eclipse.jdt.internal.debug.core.model.LambdaUtils.isLambdaFrame(jdiFrame)) {
-				IEditorInput editorInput = editor.getEditorInput();
-				IDocumentProvider provider = textEditor.getDocumentProvider();
-				IDocument document = provider.getDocument(editorInput);
-				if (document != null && JavaUI.getEditorInputJavaElement(editorInput) != null) {
+			if (org.eclipse.jdt.internal.debug.core.model.LambdaUtils.isLambdaFrame(jdiFrame)) {
+				IEditorInput editorInput = editorPart.getEditorInput();
+				if (JavaUI.getEditorInputJavaElement(editorInput) != null) {
 					IRegion region = document.getLineInformation(jdiFrame.getLineNumber() - 1);
 					List<LambdaExpression> inLineLambdas = ToggleBreakpointAdapter.findLambdaExpressions(textEditor, region);
 					for (LambdaExpression exp : inLineLambdas) {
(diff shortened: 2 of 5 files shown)


SougandhS (commented):
Image

These load instructions will be skipped (wont be suspended) when `IJDIPreferencesConstants.PREF_STATEMENT_LEVEL_STEPPING` is enabled. - Can you confirm it ?

SougandhS:
>if that should be fixed, I guess this would make sense as a distinct change to keep the PRs small)

OK for me 👍 

>Other than that, it can step over individual bytecode instructions if no debugging information (like line numbers) is present:

Sounds nice

>If wanted, I can also try to add some tests for things like stepping without debugging information.

That would be good. please add few more tests if possible

danthe1st:
The tests failed with Jenkins because `AutomatedSuite` didn't run them in the UI thread. For now, I created a new `TestSuite` for this. What is the recommended approach for tests that should run fully in the UI thread?
Should I also move the test class to a different package?

trancexpress:
Odd, this seems to want the UI thread:

```
➜  org.eclipse.jdt.debug.tests git:(master) git grep useUIThread
pom.xml:          <useUIThread>true</useUIThread>
```

The shared launch also specifies running in the UI thread.

What is the error message from the test?

danthe1st:
> What is the error message from the test?

```
org.eclipse.ui.PartInitException: No active workbench page
	at org.eclipse.jdt.internal.ui.javaeditor.EditorUtility.openInEditor(EditorUtility.java:388)
	at org.eclipse.jdt.internal.ui.javaeditor.EditorUtility.openInSpecificEditor(EditorUtility.java:191)
	at org.eclipse.jdt.internal.ui.javaeditor.EditorUtility.openInEditor(EditorUtility.java:169)
	at org.eclipse.jdt.internal.ui.javaeditor.EditorUtility.openInEditor(EditorUtility.java:154)
	at org.eclipse.jdt.debug.tests.sourcelookup.ClassFileEditorHighlightingTest.testConstructorInPackage(ClassFileEditorHighlightingTest.java:99)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at junit.framework.TestCase.runTest(TestCase.java:177)
	at junit.framework.TestCase.runBare(TestCase.java:142)
	at org.eclipse.jdt.debug.tests.AbstractDebugTest.runBare(AbstractDebugTest.java:2993)
	at junit.framework.TestResult$1.protect(TestResult.java:122)
	at junit.framework.TestResult.runProtected(TestResult.java:142)
	at junit.framework.TestResult.run(TestResult.java:125)
	at junit.framework.TestCase.run(TestCase.java:130)
	at junit.framework.TestSuite.runTest(TestSuite.java:241)
	at junit.framework.TestSuite.run(TestSuite.java:236)
	at junit.framework.TestSuite.runTest(TestSuite.java:241)
	at org.eclipse.jdt.debug.tests.DebugSuite$1.run(DebugSuite.java:61)
	at java.base/java.lang.Thread.run(Thread.java:1516)
```
See https://github.com/eclipse-jdt/eclipse.jdt.debug/runs/70316773248

`AutomatedSuite` extends `DebugSuite` which runs all tests in a non-UI thread.
```java
/**
 * Debug test suite framework that runs test in a non UI thread.
 */
public abstract class DebugSuite extends TestSuite {
```

I could also move the test to a different test suite but I don't know which one would be better.

trancexpress:
I'm not sure, maybe the non-UI thread choice was made due t o `EventWaiter` not processing UI events while waiting... `DebugEventWaiter` on the other hand does process UI events...

From my POV no need to overcomplicate things for one test case, what is in the PR right now seems OK. If we find that the UI thread test suite is growing, we can have two top level test suites, where the UI one doesn't run tests in a separate thread.

danthe1st:
I think there's also `EvalTestSuite` which apparently does some UI tests but I don't know about that one and running those locally didn't seem to work properly for me (and I'm not sure whether that runs the tests fully in the UI thread or just does some additional processing for handling UI events).

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaStackFrameSourceDisplayAdapter.java
```diff
@@ -0,0 +1,159 @@
...
+import org.eclipse.jface.viewers.IStructuredSelection;
+import org.eclipse.ui.IEditorInput;
+import org.eclipse.ui.IEditorPart;
+import org.eclipse.ui.IWorkbenchPage;
+import org.eclipse.ui.PartInitException;
+import org.eclipse.ui.texteditor.IDocumentProvider;
+import org.eclipse.ui.texteditor.ITextEditor;
+
+/**
+ * @since 3.2
```
trancexpress: Should be possible to omit `@since`, since this is an internal package.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/LambdaStackFrameSourceDisplayAdapter.java
```diff
@@ -1,73 +0,0 @@
-/*******************************************************************************
```
trancexpress: @SougandhS are there any tests for this adapter? If not, how to test that nothing breaks when replacing it?

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaDebugShowInAdapterFactory.java
```diff
@@ -54,13 +54,8 @@ public <T> T getAdapter(Object adaptableObject, Class<T> adapterType) {
...
 			}
-			try {
-				if (adaptableObject instanceof JDIStackFrame jdiFrame
-						&& org.eclipse.jdt.internal.debug.core.model.LambdaUtils.isLambdaFrame(jdiFrame)) {
-					return (T) new LambdaStackFrameSourceDisplayAdapter();
-				}
-			} catch (DebugException e) {
-				DebugUIPlugin.log(e);
+			if (adaptableObject instanceof JDIStackFrame) {
+				return (T) javaStackFrameSourceDisplayAdapter;// ensure the same instance is reused
```
trancexpress: Please add to the comment why it should be the same instance.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaStackFrameSourceDisplayAdapter.java
```diff
@@ -0,0 +1,159 @@
...
+				if (event.getContext() instanceof IStructuredSelection selection && selection.getFirstElement() instanceof JDIStackFrame frame
+						&& frame.equals(currentClassFileFrame)) {
+					return;
+				}
+
+				if (currentClassFileEditor != null) {
+					currentClassFileEditor.unhighlight();
+					currentClassFileEditor = null;
+					currentClassFileFrame = null;
+					service.removeDebugContextListener(this);
```
trancexpress: Can you add a test that debugs, then terminates the launch, then checks that we didn't leak a listener here?

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaStackFrameSourceDisplayAdapter.java
```diff
@@ -0,0 +1,159 @@
...
+				}
+			}
+		}
+		return false;
+	}
+
+	private boolean openClassFileAndHighlightStackFrame(JDIStackFrame jdiFrame, IClassFileEditorInput input) throws JavaModelException, PartInitException, DebugException {
+		IJavaElement je = JavaUI.getEditorInputJavaElement(input);
+		if (je != null) {
+			IEditorPart part = JavaUI.openInEditor(je);
```
trancexpress: Can you check what happens if this preference is off?

https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.jdt.doc.user%2Freference%2Fpreferences%2Frun-debug%2Fref-run_debug.htm

```
 Reuse editor when displaying source code

The debugger displays source code in an editor when stepping through an application. When this option is on, the debugger will reuse the editor that it opened to display source from different source files. This prevents the debugger from opening an excessive number of editors. When this option is off, the debugger will open a new editor for each source file that needs to be displayed.
```

Allegedly `org.eclipse.debug.internal.ui.sourcelookup.SourceLookupFacility.display(ISourceLookupResult, IWorkbenchPage)` honors this preference via: `org.eclipse.debug.internal.ui.sourcelookup.SourceLookupFacility.openEditor(ISourceLookupResult, IWorkbenchPage)`

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaStackFrameSourceDisplayAdapter.java
```diff
@@ -0,0 +1,159 @@
...
+			if (part instanceof ClassFileEditor editor && editor.getDocumentProvider().getDocument(editor.getEditorInput()).getLength() == 0) {
```
trancexpress: Can you add a comment that explains what this check does?

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaStackFrameSourceDisplayAdapter.java
```diff
@@ -0,0 +1,159 @@
...
+					currentClassFileEditor = null;
+					currentClassFileFrame = null;
+					service.removeDebugContextListener(this);
+					classFileUnhighlighterRegistered = false;
+				}
+			}
+		}
+	}
+
+	private boolean classFileUnhighlighterRegistered = false;
```
trancexpress: I believe this is checked in the UI thread, but set from a non-UI thread? I don't think debug events are propagated in the UI thread at least and display source sounds like UI thread code.

Can you debug and check? Best to use `AtomicBoolean` if that is the case. Or maybe `volatile` is enough.

I'm also not sure what happens if a debug event unregisters the listener concurrently to the code checking it it should be registered.

I'm also not sure about `displaySource` just running the UI code, e.g. `org.eclipse.debug.internal.ui.elements.adapters.StackFrameSourceDisplayAdapter.displaySource(Object, IWorkbenchPage, boolean)` schedules a job to do that...

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/LambdaStackFrameSourceDisplayAdapter.java
danthe1st: I (more or less) renamed/generalized this class to `JavaStackFrameSourceDisplayAdapter` and added the logic for `ClassFileEditor`.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaStackFrameSourceDisplayAdapter.java
danthe1st: My implementation opens one editor per file even if the preference is enabled (I think it's the same for the lambda-based implementation). I'll take a look but I don't know whether I can reuse the existing logic (it might be necessary to reimplement the logic or make `SourceLookupFactory#display` return the `IEditorPart`).

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaStackFrameSourceDisplayAdapter.java
danthe1st: I think it is called in a UI thread, at least in my testing.
image

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaStackFrameSourceDisplayAdapter.java
danthe1st: Done in 4a6b692dc9fde1965f50a7e288db576ad920adae.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaStackFrameSourceDisplayAdapter.java
danthe1st: Done in 4a6b692dc9fde1965f50a7e288db576ad920adae using the new `openInEditor` method which uses `SourceLookupFactory#display`.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaDebugShowInAdapterFactory.java
danthe1st: This is the case because `T` is `ISourceDisplay` (checked by the `if`) which is the type of `javaStackFrameSourceDisplayAdapter` (I now changed that field to use the interface as well). I hope this is clear now.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaStackFrameSourceDisplayAdapter.java
danthe1st: I kept that when moving it from `LambdaStackFrameSourceDisplayAdapter`. Should I still remove it?

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaStackFrameSourceDisplayAdapter.java
danthe1st: Is there a good way to check that the listener has been removed or should I just add a getter method for `classFileUnhighlighterRegistered` that's just there for the test?

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/LambdaStackFrameSourceDisplayAdapter.java
SougandhS: @trancexpress Tests weren't added, it simply highlights the inline lambda exp on lambda breakpoint suspension 

see https://eclipse.dev/eclipse/markdown/?f=news/4.39/jdt.md#improved-highlighting-for-inline-chained-lambdas

```
import java.util.Arrays;
import java.util.List;
import java.util.Optional;

public class BreakpointsTest {
	
		callMe();
	}

	private static void callMe() { 
		BreakpointsTest t = new BreakpointsTest();
		t.lotOfLambdas();
	}
	
	void lotOfLambdas() {
		Optional.of("aa").map(this::a6).filter(x -> x.isBlank()).map(this::a6);

		List<String> list = Arrays.asList("a", "b", "c");
                // create a bp for any chained exp and see highlighting on suspension 
		list.stream().filter(s -> s.length() > 0).map(s -> s.toLowerCase()).filter(x -> x.charAt(0) > 1).map(s -> s.toUpperCase()).filter(k -> k.indexOf(0) < 0).map(s -> s.toLowerCase()).filter(p -> !p.isBlank()).forEach(System.out::println);
		
	}
	
	String a6(String s) {
		return s;
	}
}

```

SougandhS: I had my testing with this patch, it was working fine 👍

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaStackFrameSourceDisplayAdapter.java
trancexpress: Yes please, only needed for API.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaStackFrameSourceDisplayAdapter.java
trancexpress: Unfortunately there is no access to `org.eclipse.debug.internal.ui.contexts.DebugWindowContextService.fListenersByPartId`... A method just for the test probably won't be enough, since the field can be overwritten and so the implementation can leak a listener without the test noticing.

Lets skip that.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaStackFrameSourceDisplayAdapter.java
danthe1st: Done in 54fa9161eff690f018105650166fd236a455db4c.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaDebugShowInAdapterFactory.java
trancexpress: Sorry but I still don't understand the comment with the re-use. Is it for efficiency, not creating the object many times? Or there is some state that is important?

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaStackFrameSourceDisplayAdapter.java
trancexpress: I went through `StackFrameSourceDisplayAdapter` again, it looks like the job is started to not run the source look up in the UI thread.

So likely this is a problem already with: https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/812

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/sourcelookup/ClassFileEditorHighlightingTest.java
```diff
@@ -0,0 +1,235 @@
...
+		JavaStackFrameSourceDisplayAdapter display = new JavaStackFrameSourceDisplayAdapter();
+
+		IJavaThread thread = null;
+		try {
+			thread = launchToBreakpoint(CLASS_NAME);
+
+			ClassFileEditor editor = (ClassFileEditor) EditorUtility.openInEditor(javaProject.getProject().getFile("bin/" + CLASS_NAME + ".class"));
+			IClassFileEditorInput editorInput = (IClassFileEditorInput) editor.getEditorInput();
+			IClassFile classFile = editorInput.getClassFile();
+			thread.getTopStackFrame().getLaunch().setSourceLocator(stackFrame -> classFile);
```
trancexpress: Why is this necessary?

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaDebugShowInAdapterFactory.java
danthe1st: This is because it stores state, specifically information about/for the listener ensuring that the highlighting is cleaned up when the instruction pointer is no longer at the class file.

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/sourcelookup/ClassFileEditorHighlightingTest.java
danthe1st: This makes sure it actually uses the `ClassFileEditor` for displaying the source.
This is necessary because the source file is present in the project and otherwise not show the bytecode.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/sourcelookup/JavaStackFrameSourceDisplayAdapter.java
danthe1st: I switched to using `synchronized` in b07279bcc503b3dc180b464683e33c89cbe81ab6.

The reason for using `synchronized` over `volatile`/atomic variables is that there are multiple variables involved and if something tries to call `displaySource()` concurrently for whatever reason (I don't think this should happen under normal circumstances but I don't know for sure), it should still ensure that these variables are updated together and there's no hard to debug/reproduce in-between state.

trancexpress:
I think https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/812 went in the wrong direction and this PR follows with more code.

With the addition of `LambdaStackFrameSourceDisplayAdapter`, every time I do debug steps inside lambda code, there is a source lookup call in the UI thread: `SourceLookupFacility.lookup()`

With this PR adding on top, a source lookup is done on every debug step - in the UI thread.

Normally the chain is: `SourceLookupFacility.displaySource()` -> `SourceLookupJob` (non-UI) -> `SourceDisplayJob` (UI) -> `SourceLookupFacility.display()` (opens editors, selects correct code)

The goal here must have been to not block the UI thread with source lookups - highlighting the current stack frame in the editor is secondary to smooth UI experience. I can imagine the source lookup causing noticeable mini-freezes in big workspaces (the effect would be tickets like https://github.com/eclipse-platform/eclipse.platform/issues/2615).

Since the source lookup and display chain is skipped, custom code is added in `LambdaStackFrameSourceDisplayAdapter` and was added in this PR. Such code then needs to duplicate preference handling and such (missing in `LambdaStackFrameSourceDisplayAdapter` but added in this PR).

@SougandhS IMO we need to clean this up, this PR is a good opportunity to do so - since even more intricate code must run once the editor is opened. I need to think on how we want to operate on the editor (given the stack frame), once its opened in `SourceLookupFacility.display()`... it would be great if you can suggest ideas. Maybe we are in the wrong adapter code, I'm not sure. Maybe we need something new. Maybe this: `org.eclipse.debug.ui.IDebugEditorPresentation.addAnnotations(IEditorPart, IStackFrame)`

```
	/**
	 * Positions and adds annotations to the given editor for the specified
	 * stack frame and returns whether any annotations were added. When
	 * <code>true</code> is returned, a call will be made to remove annotations
	 * when the source selection is cleared for the stack frame. When
	 * <code>false</code> is returned, the debugger will position and add
	 * standard annotations to the editor, and a corresponding call to remove
	 * annotations will not be made. This method is called when the debugger is
	 * has opened an editor to display source for the given stack frame.
	 *
	 * @param editorPart the editor the debugger has opened
	 * @param frame the stack frame for which the debugger is displaying
	 *  source
	 * @return <code>true</code> if annotations were added to the given editor part <code>false</code> otherwise
	 */
	boolean addAnnotations(IEditorPart editorPart, IStackFrame frame);
```

Snippet where source lookups are called on every step inside lambda:

```
		int n = 3;
		int m = 20;
		Runnable r = () -> {
		for (int i = 0; i < n; ++i) {
			String l = "0".repeat(m);
			System.out.println(l);
		}
		};
		r.run();
	}
```

Once within the `Runnable.run()` body the lambda adapter runs its lookups.

SougandhS:
> I think #812 went in the wrong direction and this PR follows with more code.

> @SougandhS IMO we need to clean this up, this PR is a good opportunity to do so - since even more intricate code must run once the editor is opened. I need to think on how we want to operate on the editor (given the stack frame), once its opened in `SourceLookupFacility.display()`... it would be great if you can suggest ideas. Maybe we are in the wrong adapter code, I'm not sure. Maybe we need something new. Maybe this: `org.eclipse.debug.ui.IDebugEditorPresentation.addAnnotations(IEditorPart, IStackFrame)`

Handled this in a new [PR](https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/933) 👍

danthe1st:
> Handled this in a new [PR](https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/933) 👍

I guess my PR should wait for that and I can update it once that's merged. Hence, I'm marking this as a draft for now.

trancexpress:
@danthe1st please hold off rebasing, I'll try to adjust the test and I'll rebase in the process.

danthe1st:
> please hold off rebasing, I'll try to adjust the test and I'll rebase in the process.

Sorry, I didn't see that until now/I only saw the notification of the other PR being merged

trancexpress:
I've adjusted two things in the PR:

1. The tests no longer open the editor manually, IMO we want to test what the user will be seeing. This slows the tests somewhat, also due to creating the test project for each test. `testDisplaySourceWithClassFileEditorHighlightsLine` also compiles with debug line information, since otherwise Eclipse doesn't allow to create the line breakpoint.
2. The editor unhighlight code now runs in a UI job - I'm not sure the code belongs in the debug event notification thread, so I've moved it. Listener registration and deregistration is also done in the UI thread so no need to synchronize it anymore. @danthe1st please test if the unhighlight and listener deregistration works as expected, if this is not covered by the tests.

When I run the tests with coverage, `JavaStackFrameEditorPresenter.removeAnnotations(IEditorPart, IThread)` is not covered. I'm not sure when this code runs, we'll have to check. I'm also not sure if this code needs to unregister the listener.

I've also adjusted the test to have to run only a small part of the code in the UI thread, so I've removed the UI debug suite.

danthe1st:
> I'm not sure when this code runs, we'll have to check. I'm also not sure if this code needs to unregister the listener.

When I tested it, it didn't seem to be executed at all.

trancexpress:
> When I tested it, it didn't seem to be executed at all.

I'll try to find where its called.

One more thing to check:

What happens if I close the class editor and another editor is opened when stepping?  Will the unhighlighting work? Or if debugging multiple programs. Or the same program with multiple class files.

For the listener probably with different workbench windows we will have a problem too. Likely we need a list of registered listeners / highlighted editors, or some other mechanism that doesn't rely on fields (e.g. a task queue for the clean-up job). For the editors we likely must check if the editor is still opened before removing highlights.

I'm also not 100% sure what the life-cycle of `org.eclipse.jdt.internal.debug.ui.JDIModelPresentation` is.

Probably this is another case where we should check what platform debug is doing to remove line highlighting when the debug launch is terminated.

And it will be good to have a test for terminating the debug launch and checking that the active class editor has no highlight.

danthe1st:
> What happens if I close the class editor and another editor is opened when stepping? Will the unhighlighting work? Or if debugging multiple programs. Or the same program with multiple class files.

When I tested with what I tried yesterday (a listener that was registered using the `IWorkbenchWindow` of the editor), unhighlighting worked. I can test it again later.

> For the listener probably with different workbench windows we will have a problem too.

I can check that as well.

trancexpress:
@danthe1st if you have time, can you first check where the Java editor line highlight is removed when terminating a Java debug launch? And what the call stacks are to get to there?

danthe1st:
If `addAnnotations` returns `true`, `removeAnnotations` will be called when the instruction pointer doesn't need to be shown any more (this includes the launch being terminated). Both methods are called in a UI thread. I will update my PR accordingly.

danthe1st:
I now used `removeAnnotations` for the unhighlighting and I added an assertion that ensures this is actually done after termination. However, I think I noticed an NPE in that class so I'll fix that as well.

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/sourcelookup/ClassFileEditorHighlightingTest.java
```diff
@@ -0,0 +1,327 @@
...
+			thread = launchToBreakpoint(config);
+
+			stepOver((IJavaStackFrame) thread.getTopStackFrame());
+			expectHighlightedText("     8  getstatic java.lang.System.out : java.io.PrintStream [13]", TIMEOUT);
+
+			stepOver((IJavaStackFrame) thread.getTopStackFrame());
+			expectHighlightedText("    17  return", TIMEOUT);
+		} finally {
+			terminateAndRemove(thread);
+			expectNothingHighlighted(TIMEOUT);
```
trancexpress: This doesn't work here, unfortunately. Since if it fails, we'll leak the project (and maybe the breakpoints) to other tests.

I think you can add explicit terminating to the `try` block and checking for no highlight, and then leave the `finally` block as is. The terminate in `finally` is just in case, it doesn't hurt to terminate before that, in the actual test.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JavaStackFrameEditorPresenter.java
```diff
@@ -66,7 +78,10 @@ public boolean addAnnotations(IEditorPart editor, IStackFrame frame) {
 
 	@Override
 	public void removeAnnotations(IEditorPart editorPart, IThread thread) {
-		// nothing to clean up
+		if (editorPart instanceof ClassFileEditor classFileEditor
+				&& classFileEditor.getDocumentProvider().getDocument(editorPart.getEditorInput()).getLength() == 0) {
+			classFileEditor.unhighlight();
```
trancexpress, line 88: Yes, this looks very good now!

trancexpress:
Just the test issue with checking for no highlighting in `finally`, then we can let Copilot take a final look and I think we can merge.

danthe1st:
I fixed the aforementioned NPE which was caused by `getDocumentProvider()` returning `null` in `removeAnnotations()` if the editor was disposed (e.g. if you close the editor and then resume) and addressed your comment in 356c27c31c024c715b34835fb31a5a6621806bc5.

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/sourcelookup/ClassFileEditorHighlightingTest.java
danthe1st: Done in 356c27c31c024c715b34835fb31a5a6621806bc5.

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/sourcelookup/ClassFileEditorHighlightingTest.java
```diff
@@ -0,0 +1,361 @@
...
+		StyledText noSourceTextWidget = null;
+		ClassFileEditor editor = (ClassFileEditor) getActivePage().getActiveEditor();
+		if (editor != null) {
+			noSourceTextWidget = editor.getNoSourceTextWidget();
+		}
+		if (noSourceTextWidget != null) {
+			StyleRange[] styleRanges = noSourceTextWidget.getStyleRanges();
+			StyledTextContent content = noSourceTextWidget.getContent();
+			for (int i = 0; i < styleRanges.length; ++i) {
+				String highlightedText = content.getTextRange(styleRanges[0].start, styleRanges[0].length);
```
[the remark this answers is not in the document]
trancexpress: > styleRanges[0] instead of styleRanges[i]

Oops.

org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIStackFrame.java
```diff
@@ -1756,6 +1756,11 @@ private boolean isDifferentLambdaContext(IStackFrame frame) throws DebugExceptio
 			return true;
 		}
 		return false;
+	}
 
+	public long getCodeIndex() {
+		synchronized (fThread) {
+			return fLocation.codeIndex();
```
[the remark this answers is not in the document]
danthe1st: Is there really a point from doing that for `codeIndex()`? From what I know, this is always just a field and cannot be unavailable (in contrast to e.g. line numbers).

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/sourcelookup/ClassFileEditorHighlightingTest.java
```diff
@@ -0,0 +1,361 @@
...
+	private static void compileWithJavac(String className, String source, List<String> compilerOptions) {
+		JavaFileObject fileObject = new SourceJavaFileObject(className, source);
+		JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
+		Assume.assumeNotNull(compiler);
+
+		StandardJavaFileManager fileManager = compiler.getStandardFileManager(null, Locale.ROOT, StandardCharsets.UTF_8);
+		DiagnosticCollector<JavaFileObject> collector = new DiagnosticCollector<>();
+		CompilationTask task = compiler.getTask(null, fileManager, collector, compilerOptions, null, List.of(fileObject));
+		Boolean result = task.call();
+		assertTrue(String.valueOf(collector.getDiagnostics()), result);
```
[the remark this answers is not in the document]
danthe1st: Done in 987dba925f2a23a66244364fe179ea6417ef8af7

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/sourcelookup/ClassFileEditorHighlightingTest.java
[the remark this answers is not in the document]
danthe1st: Replaced with for-each in 987dba925f2a23a66244364fe179ea6417ef8af7.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JavaStackFrameEditorPresenter.java
```diff
@@ -66,7 +83,16 @@ public boolean addAnnotations(IEditorPart editor, IStackFrame frame) {
...
+		if (!(editorPart instanceof ClassFileEditor classFileEditor)) {
+			return;
+		}
+		IDocumentProvider documentProvider = classFileEditor.getDocumentProvider();
+		if (documentProvider == null) {
+			return;
+		}
+		if (documentProvider.getDocument(editorPart.getEditorInput()).getLength() == 0) {
+			classFileEditor.unhighlight();
+		}
```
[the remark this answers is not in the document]
danthe1st: Done in 987dba925f2a23a66244364fe179ea6417ef8af7.

org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JavaStackFrameEditorPresenter.java
```diff
@@ -58,6 +64,17 @@ public boolean addAnnotations(IEditorPart editor, IStackFrame frame) {
...
 			}
+			if (editorPart instanceof ClassFileEditor classFileEditor
+					&& documentProvider.getDocument(editorPart.getEditorInput()).getLength() == 0) {
+				String methodName = jdiFrame.getMethodName();
+				if (jdiFrame.isConstructor()) {
+					methodName = jdiFrame.getDeclaringTypeName();
+					methodName = methodName.substring(methodName.lastIndexOf('.') + 1);
+				}
+
+				classFileEditor.highlightInstruction(methodName, jdiFrame.getSignature(), jdiFrame.getCodeIndex());
```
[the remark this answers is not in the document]
danthe1st: Done in 987dba925f2a23a66244364fe179ea6417ef8af7.

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/sourcelookup/ClassFileEditorHighlightingTest.java
```diff
@@ -0,0 +1,361 @@
...
+			List<IStatus> errors = new ArrayList<>();
+			Platform.addLogListener(new ILogListener() {
+				@Override
+				public void logging(IStatus status, String plugin) {
+					if (status.matches(IStatus.ERROR)) {
+						errors.add(status);
+					}
+				}
+			});
+			resumeAndExit(thread);
```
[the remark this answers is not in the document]
danthe1st: Done in 070e8dcdc57ceb77db6e38fda93d545f27b8cb37.

org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIStackFrame.java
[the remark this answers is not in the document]
trancexpress: I think we don't need the try-catch with the current state.

trancexpress (approved):

trancexpress:
Thank you @danthe1st !

SougandhS:
@danthe1st please add an N&N entry for this feature 👍

danthe1st:
> please add an N&N entry for this feature 👍

I created https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/524.

PR #855, opened 2026-02-11
Inline o.e.jdt.launching.macosx into jdt.launching

### LINKED ISSUE #849: Inline o.e.jdt.launching.macosx
Currently the handling of `VMInstalls` for Mac, which have a special layout, is separated into the `org.eclipse.jdt.launching.macosx` bundle.
But as encountered repeatedly, this separation can cause difficulties to assemble a proper runtime, e.g. for tests since only the `org.eclipse.jdt` feature includes it and nothing else requires it.
Looking at the code and dependencies I see nothing that would cause harm on other platforms than Mac and all (in-directed) calls to it are already guarded by an OS check. And since there are only a few classes I wonder:
Can the `org.eclipse.jdt.launching.macosx` bundle be in-lined into `org.eclipse.jdt.launching`?

And similarly could `org.eclipse.jdt.launching.ui.macosx` be inlined too? Maybe into `org.eclipse.jdt.debug.ui`?

Maybe @stephan-herrmann or @iloveeclipse can answer this?

If there's no compelling reason speaking against it, I can provide a PR to move the content.

HannesWell:
## What it does
This simplifies the assembly of Eclipse runtimes on MacOS.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/849

## Author checklist


HannesWell:
@iloveeclipse what do you think about this effort respectively can you tell who else would be interested?

HannesWell:
@iloveeclipse any update on this? 
Otherwise, who else would be interested in reviewing this?

HannesWell:
@iloveeclipse can we proceed with this? I think now would be a good time.
Since neither I nor you (as far as I know) are on Mac, @BeckerWdf can you (or one of your colleagues) or @HeikoKlare help with testing this on a real Mac?

iloveeclipse:
@SougandhS : I believe you are using Mac? Can you test this patch?

SougandhS:
Hi @HannesWell 
With this patch 2 MacOS X VM is shown in `Add new JRE` page

image


EDIT: 
I tested it by simply running a child eclipse with your changes. Is it to be checked in a different way ?

HannesWell:
> With this patch 2 MacOS X VM is shown in `Add new JRE` page

Thank you for testing this!
I believe this is because you still have the to be removed `org.eclipse.jdt.launching.macosx` fragment in the launched runtime.
In general this change also requires to remove that fragment from the JDT feature:
- https://github.com/eclipse-jdt/eclipse.jdt/pull/160

I have now added a guard to prevent adding the macOS VM type twice. I think that's still useful in case some installations don't have the fragment properly removed.
It would be great if you could test this again, with and without the `org.eclipse.jdt.launching.macosx` fragment being removed from the JDT feature to also verify that the mentioned guard works as expected.

SougandhS:
I tested again with and without https://github.com/eclipse-jdt/eclipse.jdt/pull/160 and it is now not showing twice in both cases.

SougandhS (approved):
LGTM 👍

iloveeclipse:
> I tested again with and without [eclipse-jdt/eclipse.jdt#160](https://github.com/eclipse-jdt/eclipse.jdt/pull/160) and it is now not showing twice in both cases.

Thanks! OK, let us merge this.

HannesWell:
Great! Thanks to all of you.

Issue #782, opened 2025-09-19
4 Tests failing since I20250917-0250 on ep438I-unit-macosx-aarch64-java21 macosx.cocoa.aarch64 21


SougandhS:
See 
https://download.eclipse.org/eclipse/downloads/drops4/I20250918-0350/testresults/html/org.eclipse.jdt.debug.tests_ep438I-unit-macosx-aarch64-java21_macosx.cocoa.aarch64_21.html

```
testLimitModules_release9

testLimitModules1

testVeryLongClasspathWithArgumentFile

testVeryLongClasspathWithClasspathOnlyJar
```

SougandhS:
I tried the tests on I20250918-0350 with Mac aarch64 and Java 21
all tests are passing 

<img width="430" height="237" alt="Image" src="https://github.com/user-attachments/assets/718631de-c2e4-4acc-805b-344790009d7e" />

<img width="472" height="305" alt="Image" src="https://github.com/user-attachments/assets/12fa7349-3bac-435e-9d38-2986a9eedec4" />

subyssurendran666:
I have also tested the unit testes on I20250918-0350 with Mac aarch64 and Java 21. All the tests are failing

<img width="479" height="277" alt="Image" src="https://github.com/user-attachments/assets/4194568c-32e2-424f-ada3-adf525a2787d" />

<img width="465" height="260" alt="Image" src="https://github.com/user-attachments/assets/025e4a66-d58d-4808-bedb-0cdf94c1e2ab" />

iloveeclipse:
> The test seems to be finding Java 11 or below modules (Nashorn). That is not expected.

That was also my first impression looking at your modules diff. Nashorn is removed from JDK since long time, I remember all the troubles some of our customers had because of that.

iloveeclipse:
The log file is this: https://download.eclipse.org/eclipse/downloads/drops4/I20260218-1800/testresults/ep439I-unit-macosx-x86_64-java21_macosx.cocoa.x86_64_21/org.eclipse.jdt.debug.tests.AutomatedSuite.txt

trancexpress:
> [@trancexpress](https://github.com/trancexpress) : assuming the default JRE is switched somewhere by some test before, shouldn't we just set **latest** available JRE explicitly in this test? Or whatever JRE version is expected.

[...]

PR #864, opened 2026-02-18
Add more logging to tests, fail test if project has JVM version below 21

trancexpress:
See: #782


org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/ModuleOptionsTests.java
```diff
@@ -178,8 +181,10 @@ public void testLimitModules_release9() throws CoreException {
 		}
 	}
 
-	public void testLimitModules1() throws JavaModelException {
+	public void testLimitModules1() throws Exception {
 		IJavaProject javaProject = getProjectContext();
+		javaProject.setOption(JavaCore.COMPILER_RELEASE, JavaCore.DISABLED);
```
iloveeclipse: Why is this needed?

iloveeclipse (approved):

iloveeclipse:
> I don't think we are asserting the right thing at the moment.

The one is classpath setup for the project by the test, the other is VM actually used. So the assert is correct. We don't want to talk with Java 11 VM, it will never have expected modules from Java 21 :-).

PR #873, opened 2026-02-20
Narrow down scheduling rule in JavaDebugOptionsManager.updateBreakpointMessages()

### LINKED ISSUE #872: Narrow down scheduling rule in JavaDebugOptionsManager.updateBreakpointMessages()
See suggestion by @mickaelistria here: https://github.com/eclipse-platform/eclipse.platform/pull/2218#issuecomment-3466583210

The same suggestion can be applied to fix: https://bugs.eclipse.org/bugs/show_bug.cgi?id=540132

Since its better to lock less than to lock more, we should apply this suggestion also to: `org.eclipse.jdt.internal.debug.ui.JavaDebugOptionsManager.updateBreakpointMessages(IBreakpoint[])`

trancexpress:
The fix for https://bugs.eclipse.org/bugs/show_bug.cgi?id=540132 introduced a workspace modification rule, when `JavaDebugOptionsManager.updateBreakpointMessages()` is running.

This change narrows down this rule to a `MultiRule` for the involved markers.

Fixes: #872

## Author checklist

### DIFF
--- a/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JavaDebugOptionsManager.java
+++ b/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/JavaDebugOptionsManager.java
@@ -16,28 +16,27 @@ package org.eclipse.jdt.internal.debug.ui;
 
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.StringTokenizer;
 
 import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.resources.IMarkerDelta;
 import org.eclipse.core.resources.IResource;
-import org.eclipse.core.resources.IResourceRuleFactory;
-import org.eclipse.core.resources.IWorkspace;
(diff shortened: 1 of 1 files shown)


trancexpress:
> See discussion on https://bugs.eclipse.org/bugs/show_bug.cgi?id=66538. Somehow I believe we are missing something.

https://bugs.eclipse.org/bugs/show_bug.cgi?id=66538#c21https://bugs.eclipse.org/bugs/show_bug.cgi?id=66538#c21


> I believe we are compliant, taking into consideration John A's last comment. 
We simply ask the workspace to execute a runnable. It has no "outer" 
scheduling rule, but loops thru deletions on the insdie (i.e. deleting 
individual markers). Each marker deletion is responsible for getting a lock as 
required. We can't pass in effective progress monitors as our API for deleting 
breakpoints does not accept progress monitors.
>
> The net result is that we are now using proper marker modification scheduling 
rules when we modify breakpoints.


I do see `Marker.delete()` uses a marker scheduling rule... but that is `null`. Maybe we should use a different rule there? I don't know...

In addition the code here will still need rules for the markers in some way, while updating the attributes. Right now this is the workspace rule we initially added, or the rule added with this PR... I'm not sure if we can lock a rule within the code, per marker. It would have to wrap the `marker.exists()` call and `marker.setAttribute()`.

Issue #851, opened 2026-02-11
[TEST] LongClassPathTests failure on Windows


SougandhS:
See https://download.eclipse.org/eclipse/downloads/drops4/I20260210-1800/testresults/html/org.eclipse.jdt.debug.tests_ep439I-unit-win32-x86_64-java21_win32.win32.x86_64_21.html

TestSuite with 2 tests [example: testVeryLongClasspathWithClasspathOnlyJar(org.eclipse.jdt.debug.tests.launching.LongClassPathTests)] | testVeryLongClasspathWithArgumentFile | Error | Program did not suspend, launch terminated.org.eclipse.jdt.debug.tests.AbstractDebugTest$TestAgainException: Program did not suspend, launch terminated.at org.eclipse.jdt.debug.tests.AbstractDebugTest.launchAndWait(AbstractDebugTest.java:1453)at org.eclipse.jdt.debug.tests.AbstractDebugTest.launchAndWait(AbstractDebugTest.java:1403)at org.eclipse.jdt.debug.tests.AbstractDebugTest.launchAndWait(AbstractDebugTest.java:1385)at org.eclipse.jdt.debug.tests.launching.LongClassPathTests.launchAndSuspend(LongClassPathTests.java:280)at org.eclipse.jdt.debug.tests.launching.LongClassPathTests.testVeryLongClasspathWithArgumentFile(LongClassPathTests.java:156)at java.base/java.lang.reflect.Method.invoke(Method.java:580)at junit.framework.TestCase.runTest(TestCase.java:177)at junit.framework.TestCase.runBare(TestCase.java:142)at org.eclipse.jdt.debug.tests.AbstractDebugTest.runBare(AbstractDebugTest.java:2929)at junit.framework.TestResult$1.protect(TestResult.java:122)at junit.framework.TestResult.runProtected(TestResult.java:142)at junit.framework.TestResult.run(TestResult.java:125)at junit.framework.TestCase.run(TestCase.java:130)at junit.framework.TestSuite.runTest(TestSuite.java:241)at junit.framework.TestSuite.run(TestSuite.java:236)at junit.framework.TestSuite.runTest(TestSuite.java:241)at org.eclipse.jdt.debug.tests.DebugSuite$1.run(DebugSuite.java:61)at java.base/java.lang.Thread.run(Thread.java:1583) | 504.598
-- | -- | -- | -- | --
[...]

PR #880, opened 2026-02-26
Set default VM for execution environment in LongClassPathTests

trancexpress:
Fixes: #851


trancexpress:
Looking at https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/851#issuecomment-3965057343, it seems test snippets are compiled with one Java version and ran on another version.

When I run the tests locally, they fail, but I don't see the unexpected class versions errors.

While the change here fixes the tests when I run them locally, I think we should understand why tests are compiled with one Java version and ran on another...

Issue #843, opened 2026-01-31
MultiReleaseLaunchTests has problems on Windows


stephan-herrmann:
We are observing MultiReleaseLaunchTests to fail with the following error in particular on Windows:
```
The following java versions are required by this test but can not be found: [RequiredJavaVersion[from=11, to=16], RequiredJavaVersion[from=17, to=20]]
```

The test uses several strategies for location necessary JRE installations, one of which is searching in `/opt/tools/java/openjdk/`. Obviously that path doesn't exist on windows.

For that reason I suggest we add another fallback: search within the parent directory of the java.home of the running JVM.

HannesWell:
> My latest change assumed that somewhere on disk there is some folder like /org/foo/java which contains subfolders for all relevant JDK installations.
> 
> OTOH, we don't want to require all JDT developers to create various environment variables.

That makes sense.
But a multi-stage approach could fit for both cases:
1. Check if `JAVA_XX_HOME` environment variables exist and point to existing folders
2. Search for co-located JVMs with a suitable heuristic.

The second point would then only be executed if these env-variables don't exist.


> > IIRC some JDT or PDE tests could use a JDK-8 but are skipped if none is available. Does this sound good?
> 
> Not sure what you are proposing here? Do you suggest the test in question here should simply be skipped if its requirements are not met?

Actually I intended the opposite: Enable that test for I-builds when these other JDKs are available.

laeubi:
> @laeubi maybe you can weigh in how your test from https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/753 can be made to succeed in all relevant environments: locally, CI-builds, and here: production builds, where it still fails on windows.

The requirement is impossible to fulfill in general due it obviously would be insufficient to scan the whole filesystem, we do not want to automatically download JVMs and all environments can be arbitrarily be configured.

Then the test _has_ worked for a longer time now and is already quite lenient about the actual JVM (e.g. it uses ranges for matching versions) and is using the JDT autodiscovery feature, so I'm not sure what else can be made here,

> My latest change assumed that somewhere on disk there is some folder like /org/foo/java which contains subfolders for all relevant JDK installations.

[...]

PR #891, opened 2026-03-04
Re-enable MultiReleaseLaunchTests on Windows and MacOS

trancexpress:
See: #843


HannesWell:
> Or do we need more changes @HannesWell, for the JVMs to be seen by the JDT debug tests?

Unfortunately yes. But I have this on my to-do list.
The test configuration site is still missing.

PR #930, opened 2026-04-02
Fix Debug Hover showing SuperClass Field's value in Subclass Field

### LINKED ISSUE #924: Debugger shows wrong value for a private field defined also in the super class
Found while reviewing https://github.com/eclipse-platform/eclipse.platform/pull/2582.

Change `ToggleStepFiltersAction` to have constructor for better reproduction and add breakpoint in the constructor, as shown in the screenshot.

After `ToggleStepFiltersAction.fInitialized` value is set to **false**, hover over the `ToggleStepFiltersAction.fInitialized` to see the value.
It shows **true** !!! which is where you start to believe in magic, until you notice that the super class `DebugCommandAction` also has private field with exact same name but different value. OMG.

Image

Obviously, debugger shows wrong velue on hover.

SougandhS:
This commit adds an additional computation for finding the subclass class field if superclass has same field as of subclass

Fixes : https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/924

## Author checklist

### DIFF
--- a/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIObjectValue.java
+++ b/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIObjectValue.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2015 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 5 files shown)


org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIObjectValue.java
```diff
@@ -277,6 +277,17 @@ public IJavaFieldVariable getField(final String name, final String declaringType
 						field = fieldTmp;
 						break;
 					}
+					List<Field> fieldList = ref.allFields(); // Possible sub class fields with same name as of super class
+					fieldList.remove(fieldTmp);
```
iloveeclipse, line 281: @SougandhS : This line will cause CME on next iteration if the break at line 287 will be not reached.

You can't modify the list while iterating it, you can only use `fields.remove()` here

SougandhS: Will raise a fix now

Issue #940, opened 2026-04-15
ExpressionEvalTest.test547462_BooleanExpression failed with "Address already in use" in I20260414-1800


trancexpress:
See: https://download.eclipse.org/eclipse/downloads/drops4/I20260414-1800/testresults/html/org.eclipse.jdt.debug.tests_ep440I-unit-macosx-x86_64-java21_macosx.cocoa.x86_64_21.html

```
	Cannot connect to VM

org.eclipse.core.runtime.CoreException: Cannot connect to VM
at org.eclipse.jdt.launching.AbstractVMRunner.abort(AbstractVMRunner.java:56)
at org.eclipse.jdt.internal.launching.StandardVMDebugger.run(StandardVMDebugger.java:457)
at org.eclipse.jdt.launching.JavaLaunchDelegate.launch(JavaLaunchDelegate.java:176)
at org.eclipse.debug.internal.core.LaunchConfiguration.launch(LaunchConfiguration.java:778)
at org.eclipse.jdt.debug.tests.AbstractDebugTest.launchAndWait(AbstractDebugTest.java:1495)
at org.eclipse.jdt.debug.tests.AbstractDebugTest.launchAndWait(AbstractDebugTest.java:1473)
at org.eclipse.jdt.debug.tests.AbstractDebugTest.launchToBreakpoint(AbstractDebugTest.java:1669)
at org.eclipse.jdt.debug.tests.AbstractDebugTest.launchToBreakpoint(AbstractDebugTest.java:1624)
at org.eclipse.jdt.debug.tests.AbstractDebugTest.launchToBreakpoint(AbstractDebugTest.java:1596)
at org.eclipse.jdt.debug.tests.AbstractDebugTest.launchToBreakpoint(AbstractDebugTest.java:1583)
at org.eclipse.jdt.debug.tests.eval.ExpressionEvalTest.debugWithBreakpoint(ExpressionEvalTest.java:116)
at org.eclipse.jdt.debug.tests.eval.ExpressionEvalTest.test547462_BooleanExpression(ExpressionEvalTest.java:66)
at java.base/java.lang.reflect.Method.invoke(Method.java:580)
at junit.framework.TestCase.runTest(TestCase.java:177)
at junit.framework.TestCase.runBare(TestCase.java:142)
at org.eclipse.jdt.debug.tests.AbstractDebugTest.runBare(AbstractDebugTest.java:3007)
[...]

PR #943, opened 2026-04-18
Improve BindException error message to include port

trancexpress:
Debug tests fail infrequently with address already in use. The `BindException` thrown in the JRE doesn't include the used port. This change wraps the `BindException` to include the port in the message.

See: https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/940

## Author checklist


trancexpress:
Copilot suggests not adding the information for port 0. I'm not sure we handle port 0, but if do, then we found no available port at all. Which is also useful information, just not presented in a nice way. So IMO not worth complicating the code.

PR #969, opened 2026-06-21
Fix VirtualThreadsDebugViewTests

stephan-herrmann:
+ ensure test project TwoThree sets also target to 23.


stephan-herrmann:
See https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/922#issuecomment-4762985318

PR #972, opened 2026-07-02
Consider "RMI Runtime" thread group as a user group

### LINKED ISSUE #971: Consider "RMI Runtime" thread group as a user group and not a system group
While debugging LSP code that is triggered by LSP4J, I've noticed that the threads which service LSP requests are considered system threads.

They are under the "RMI Runtime" thread group, defined here: https://github.com/openjdk/jdk/blob/10af769eb06427e37ae943a21964ae51de4526c6/src/java.rmi/share/classes/sun/rmi/runtime/RuntimeUtil.java#L76

```
    /**
     * Special child of the system thread group for running tasks that
     * may execute user code. The need for a separate thread group may
     * be a vestige of it having had a different security policy from
     * the system thread group, so this might no longer be necessary.
     */
    private static final ThreadGroup userThreadGroup =
        new ThreadGroup(systemThreadGroup, "RMI Runtime");
```

Threads created with this group are called user threads by the JDK: https://github.com/openjdk/jdk/blob/10af769eb06427e37ae943a21964ae51de4526c6/src/java.rmi/share/classes/sun/rmi/runtime/RuntimeUtil.java#L182

```
    public static Thread newUserThread(Runnable runnable, String name, boolean daemon) {
        return newThread(userThreadGroup, runnable, name, daemon);
    }
```

Such threads seem to be created to service TCP socket communication: https://github.com/openjdk/jdk/blob/10af769eb06427e37ae943a21964ae51de4526c6/src/java.rmi/share/classes/sun/rmi/transport/tcp/TCPTransport.java#L101

```
    private static final ExecutorService connectionThreadPool =
        new ThreadPoolExecutor(0, maxConnectionThreads,
            threadKeepAliveTime, TimeUnit.MILLISECONDS,
            new SynchronousQueue<Runnable>(),
            new ThreadFactory() {
                public Thread newThread(Runnable runnable) {
                    return RuntimeUtil.newUserThread(runnable, "TCP Connection(idle)", true);
                }
            });
```

See: https://github.com/openjdk/jdk/blob/10af769eb06427e37ae943a21964ae51de4526c6/src/java.rmi/share/classes/sun/rmi/transport/tcp/TCPTransport.java#L356-L385

```
        /**
         * Accepts connections from the server socket and executes
         * handlers for them in the thread pool.
         **/
        private void executeAcceptLoop() {
            if (tcpLog.isLoggable(Log.BRIEF)) {
                tcpLog.log(Log.BRIEF, "listening on port " +
                           getEndpoint().getPort());
            }

            while (true) {
                Socket socket = null;
                try {
                    socket = serverSocket.accept();

                    /*
                     * Find client host name (or "0.0.0.0" if unknown)
                     */
                    InetAddress clientAddr = socket.getInetAddress();
                    String clientHost = (clientAddr != null
                                         ? clientAddr.getHostAddress()
                                         : "0.0.0.0");

                    /*
                     * Execute connection handler in the thread pool,
                     * which uses non-system threads.
                     */
                    try {
                        connectionThreadPool.execute(
                            new ConnectionHandler(socket, clientHost));
```

In this context I have 2 issues with JDT debug:

1. To debug LSP code, I had to enable system threads. System threads are not shown per default so I was wondering what is going on until I thought to enable "show system threads" in the Debug view.
2. We still have this bug, which is probably (very) difficult to fix: https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/846

So my question is, should we consider threads in the group "RMI Runtime" to not be system threads? Considering JDK code calls them user threads?

Currently we have this in `JDIThread`: https://github.com/eclipse-jdt/eclipse.jdt.debug/blob/8623659cfdc3d6dbe570e810be1112566a89e703/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIThread.java#L547

```
	protected void determineIfSystemThread() throws DebugException {
		fIsSystemThread = false;
		ThreadGroupReference tgr = getUnderlyingThreadGroup();
		fIsSystemThread = tgr != null;
		while (tgr != null) {
			String tgn = null;
			try {
				tgn = tgr.name();
				tgr = tgr.parent();
			} catch (UnsupportedOperationException e) {
				fIsSystemThread = false;
				break;
			} catch (RuntimeException e) {
				targetRequestFailed(
						MessageFormat.format(
								JDIDebugModelMessages.JDIThread_exception_determining_if_system_thread,
								e.toString()), e);
				// execution will not reach this line, as
				// #targetRequestFailed will throw an exception
				return;
			}
			if (tgn != null && tgn.equals(MAIN_THREAD_GROUP)) {
				fIsSystemThread = false;
				break;
			}
		}
	}
```

The main thread group being: https://github.com/eclipse-jdt/eclipse.jdt.debug/blob/8623659cfdc3d6dbe570e810be1112566a89e703/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIThread.java#L114

```
	/**
	 * Constant for the name of the main thread group.
	 */
	private static final String MAIN_THREAD_GROUP = "main"; //$NON-NLS-1$
```

Should we add a 2nd constant for "RMI Runtime" and use that as well, to determine if a thread is a system thread?

trancexpress:
LSP4J calls are serviced by the "RMI Runtime" thread group. In JDK code, this group is considered a user thread group, as it can run user code (such as LSP4J calls from clients).

This change adjusts `JDIThread` to consider the "RMI Runtime" thread group as a user thread group. This makes it possible to debug LSP4J code without turning on the system threads Debug view filter. Which in turn avoids running into Debug view updating mechanism bugs related to using filters.

See: https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/971

### DIFF
--- a/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIThread.java
+++ b/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIThread.java
@@ -114,6 +114,12 @@ public class JDIThread extends JDIDebugElement implements IJavaThread {
 	private static final String MAIN_THREAD_GROUP = "main"; //$NON-NLS-1$
 
 	/**
+	 * Constant for the name of the {@code RMI Runtime} thread group. See internal JDK method {@code sun.rmi.runtime.RuntimeUtil.newUserThread()},
+	 * which uses: {@code sun.rmi.runtime.RuntimeUtil.userThreadGroup}
+	 */
+	private static final String RMI_RUNTIME_THREAD_GROUP = "RMI Runtime"; //$NON-NLS-1$
+
+	/**
 	 * @since 3.5
 	 */
 	public static final int RESUME_QUIET = 500;
@@ -565,7 +571,7 @@ public class JDIThread extends JDIDebugElement implements IJavaThread {
 				// #targetRequestFailed will throw an exception
 				return;
 			}
-			if (tgn != null && tgn.equals(MAIN_THREAD_GROUP)) {
+			if (MAIN_THREAD_GROUP.equals(tgn) || RMI_RUNTIME_THREAD_GROUP.equals(tgn)) {
 				fIsSystemThread = false;
 				break;
 			}
--- a/org.eclipse.jdt.debug/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.debug/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jdt.debug; singleton:=true
-Bundle-Version: 3.26.0.qualifier
+Bundle-Version: 3.26.100.qualifier
 Bundle-ClassPath: jdimodel.jar
 Bundle-Activator: org.eclipse.jdt.internal.debug.core.JDIDebugPlugin
 Bundle-Vendor: %providerName


trancexpress:
@HannesWell is there something wrong with the bot that bumps versions? Or its only running for the platform and not for JDT?

iloveeclipse:
It's also broken on platform

org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIThread.java
```diff
@@ -113,6 +113,11 @@ public class JDIThread extends JDIDebugElement implements IJavaThread {
 	 */
 	private static final String MAIN_THREAD_GROUP = "main"; //$NON-NLS-1$
 
+	/**
+	 * Constant for the name of the {@code RMI Runtime} thread group.
```
iloveeclipse: I would  mention `sun.rmi.runtime.RuntimeUtil.newUserThread(Runnable, String, boolean)` method and the used `sun.rmi.runtime.RuntimeUtil.userThreadGroup` thread group.

iloveeclipse:
I would say, let try this fix it as proposed. `sun.rmi.runtime.RuntimeUtil.userThreadGroup` is named "user thread" at the end. Please rebase & double check after if the version bump is still needed.

trancexpress:
> double check after if the version bump is still needed

It looks like its needed, there were failing checks without it.

trancexpress:
Do we proceed here @iloveeclipse ?

iloveeclipse (approved):

PR #986, opened 2026-08-27
fix #985: cache class classifications

### LINKED ISSUE #985: Debugger very slow showing variables/stack trace at breakpoint due to StackFrameCategorizer rescanning filesystem on every breakpoint hit
## Summary
When hitting a breakpoint in a large Java project, Eclipse takes 10–20+ seconds 
before variables and stack trace are displayed. The delay occurs on every 
breakpoint hit, but occasionally results are instant (cache hit).

## Environment
- Eclipse 2026-03
- OpenJDK 21.0.11+10-LTS
- Windows 11, 32 GB RAM
- Large multi-project workspace

## Root Cause (identified via thread dumps)

Thread dumps captured during the slow period consistently show `Worker-2: Debug 
child count update` stuck in `StackFrameCategorizer.categorize()` → 
`ContainerSourceContainer.findSourceElements()` → `Win32Handler.FindFirstFileW`.

The categorizer calls `findSourceElements()` **for every stack frame, on every 
breakpoint hit**, performing a recursive Windows filesystem scan to categorize 
each frame as "platform", "library" or "application" code.

In stack2.txt (timestamp 13:40:25), Worker-2 had consumed 10,609ms of CPU and 
was 10 recursive calls deep in `ContainerSourceContainer.findSourceElements()`,
scanning directories via `Win32Handler.FindFirstFileW`.

At the same time, `Worker-13: Label Job` was blocked in 
`ThreadJob.waitForRun()`, waiting for Worker-2 to finish before it could render 
the Debug view labels — which is why the user sees nothing until the scan 
completes.

## Why is it intermittently fast?
The single-entry cache in `AbstractSourceLookupDirector` occasionally produces 
a hit when the same class is encountered again. But for the initial breakpoint 
hit the cache is always cold.

## Expected Behavior
`StackFrameCategorizer` should cache results **per class name** across all 
frames in a single breakpoint suspension event, so that the same class is 
never looked up more than once per breakpoint hit. Ideally the cache should 
also persist across breakpoint hits within the same debug session and only 
be invalidated when source containers change.

## Actual Behavior
The filesystem is rescanned for every single stack frame on every breakpoint 
hit, regardless of whether the same class was already looked up moments ago.

## Workarounds tried
- Reducing source lookup path to minimum required projects → still slow
- Enabling Advanced Source Lookup → no change (categorizer bypasses it)
- Adding Windows Defender exclusions → marginal improvement only

## Additional Finding: Confirmed by disabling the categorizer

Setting `stackFrameCategorizer` to `null` in `JDIDebugPlugin` **eliminates 
the delay entirely**. The relevant code in `JDIDebugPlugin`:

```java
public synchronized IJavaStackFrame.Category getStackFrameCategorization(
        IJavaStackFrame frameToCategorize) {
    return stackFrameCategorizer != null 
        ? stackFrameCategorizer.apply(frameToCategorize) 
        : null;
}
```

The categorizer is created and registered in `JDIDebugUIPlugin`:

```java
IEclipsePreferences instancePreferences = getInstancePreferences();
stackFrameCategorizer = new StackFrameCategorizer(
        Platform.getPreferencesService(), instancePreferences);
instancePreferences.addPreferenceChangeListener(stackFrameCategorizer);
JDIDebugPlugin.getDefault().setStackFrameCategorizer(
        stackFrameCategorizer::categorize);
```

When the categorizer returns `null`, `JDIStackFrame.getCategory()` skips 
the source lookup entirely, and variables/stack trace appear **immediately**.

This confirms that `StackFrameCategorizer` is the **sole cause** of the 
delay — the debugger itself is fast, only the categorization is not.

## Attached
[stack.txt](https://github.com/user-attachments/files/31465597/stack.txt)
[stack1.txt](https://github.com/user-attachments/files/31465594/stack1.txt)
[stack2.txt](https://github.com/user-attachments/files/31465596/stack2.txt)
[stack3.txt](https://github.com/user-attachments/files/31465592/stack3.txt)
[stack4.txt](https://github.com/user-attachments/files/31465593/stack4.txt)

gzsombor:
## What it does
It caches the source file lookup so the slow windows file system checks are not repeated during debug.

## How to test

Start with debugging, evaluate step-by-step.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/StackFrameCategorizer.java
+++ b/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/StackFrameCategorizer.java
@@ -15,8 +15,10 @@ package org.eclipse.jdt.internal.debug.ui;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.StringTokenizer;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -25,9 +27,13 @@ import org.eclipse.core.runtime.preferences.IEclipsePreferences;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences.IPreferenceChangeListener;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences.PreferenceChangeEvent;
 import org.eclipse.core.runtime.preferences.IPreferencesService;
+import org.eclipse.debug.core.DebugEvent;
 import org.eclipse.debug.core.DebugException;
+import org.eclipse.debug.core.IDebugEventSetListener;
+import org.eclipse.debug.core.model.ISourceLocator;
 import org.eclipse.jdt.core.IClassFile;
 import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.jdt.debug.core.IJavaDebugTarget;
 import org.eclipse.jdt.debug.core.IJavaStackFrame;
 import org.eclipse.jdt.debug.core.IJavaStackFrame.Category;
 import org.eclipse.jdt.internal.ui.filtertable.Filter;
@@ -35,7 +41,7 @@ import org.eclipse.jdt.internal.ui.filtertable.Filter;
 /**
  * Service to help categorize the stack frames into {@link IJavaStackFrame.Category}, based on the internally stored preferences.
  */
-public class StackFrameCategorizer implements IPreferenceChangeListener {
+public class StackFrameCategorizer implements IPreferenceChangeListener, IDebugEventSetListener {
 	private final static String PREFIX = JDIDebugUIPlugin.getUniqueIdentifier() + ".enable_category_"; //$NON-NLS-1$
 
 	/**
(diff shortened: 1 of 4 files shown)


org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/StackFrameCategorizer.java
```diff
@@ -187,32 +222,68 @@ public IJavaStackFrame.Category categorize(IJavaStackFrame frame) {
...
+
+	/**
+	 * When a debug target terminates, the entries held for its (now stale) source locator are dropped, so the cache does not grow without bound
+	 * across debug sessions.
+	 */
+	@Override
+	public void handleDebugEvents(DebugEvent[] events) {
+		for (DebugEvent event : events) {
+			if (event.getKind() == DebugEvent.TERMINATE && event.getSource() instanceof IJavaDebugTarget) {
+				clearCache();
```
SougandhS: `clearCache()` calls `sourceOriginCache.clear()` which wipes entries for all launches, not just the one that terminated. If two debug sessions are running simultaneously and one ends, the entire cache gets cleared away, forcing the still-running session to redo all its expensive filesystem scans from scratch at the next breakpoint hit.

Can we do something like  ?

```
if (event.getKind() == DebugEvent.TERMINATE && event.getSource() instanceof IJavaDebugTarget target) {
    ISourceLocator locator = target.getLaunch().getSourceLocator();
    sourceOriginCache.keySet().removeIf(key -> key.sourceLocator() == locator);
}
```

gzsombor: Yes, good idea, I was more worried about accumulating unnecessary cached entries, but that shouldnt be a real issue. :crossed_fingers:

SougandhS:
@gzsombor Could you please update the commit message to follow the Eclipse commit message standards and squash all commits into a single commit?
Thanks!
image

gzsombor:
Done!

> @gzsombor Could you please update the commit message to follow the Eclipse commit message standards and squash all commits into a single commit? Thanks! image

SougandhS (approved):

SougandhS:
Thanks @gzsombor !

gzsombor:
Welcome @SougandhS !

PR #994, opened 2026-09-07
[Exp] Replace ASM with JVM Classfile API in Advanced Source Lookup agent for Java 25+

SougandhS:
Add javaagent25 jar for Class-File API based advanced source lookup. ASM must be updated for each new jdk. Debuggees on Java 25+ can use the JDK's built-in java.lang.classfile API instead, which always matches its own JDK's class format. The existing ASM-based javaagent jar is unchanged and still used for debuggees below Java 25

## Author checklist


org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/StratumTests.java
```diff
@@ -141,4 +141,8 @@ public void testGetSourceNameInStratum() throws Exception {
 			removeAllBreakpoints();
 		}
 	}
+
+	private boolean isLessThanJava25() {
+		return Runtime.version().feature() < 25;
```
iloveeclipse: This mixes debugger runtime version with the target VM versions of debuggee, it is not same!

iloveeclipse (changes requested):
If I see it right, this would restrict debuggee to Java 25+. However, we still want debug Java 8 applications.

SougandhS:
> If I see it right, this would restrict debuggee to Java 25+. However, we still want debug Java 8 applications.

it still uses old ASM based javaagent for < 25

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/StratumTests.java
SougandhS: fixed 👍

mpalat:
> > If I see it right, this would restrict debuggee to Java 25+. However, we still want debug Java 8 applications.
> 
> it still uses old ASM based javaagent for < 25

We had discussed about moving from ASM in PMC  last week - and also on a possibility of our own (JDT) Disassembler instead of ASM. 

For  a similar work @SougandhS : also please have a look at this too: https://github.com/eclipse-pde/eclipse.pde/pull/2477#issuecomment-5655563248

mpalat:
@SougandhS - as we briefly discussed offline, let us not shift to ClassFile APIs completely. Let's use the classfile APIs for 25 and keep the ASM implementation (and dependency) as it is. Let's continue this model for sometime and see how it goes and then take a call after a few releases. This middle level transition would make sure that we have minimal disruption for our customers/upline products. I would let others (@iloveeclipse ?) weigh in here as well. Thanks!

SougandhS:
> @SougandhS - as we briefly discussed offline, let us not shift to ClassFile APIs completely. Let's use the classfile APIs for 25 and keep the ASM implementation (and dependency) as it is. 

We'll create a new module for this 
eg. `org.eclipse.jdt.launching.ClassFileJavaAgent`

mpalat:
@SougandhS  - see also: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5418

SougandhS:
Hi @iloveeclipse, can you re-check this new implementation ?

org.eclipse.jdt.launching/launching/org/eclipse/jdt/launching/sourcelookup/advanced/AdvancedSourceLookup.java
```diff
@@ -107,7 +107,7 @@ private static void addWorkspaceLocation(Collection<File> locations, IPath works
 	 * Returns {@code -javaagent} jvm launch argument.
 	 */
 	public static String getJavaagentString() {
-		return AdvancedSourceLookupSupport.getJavaagentString();
+		return AdvancedSourceLookupSupport.getJavaagentLocation();
```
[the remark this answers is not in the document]
SougandhS: Done

org.eclipse.jdt.launching.javaagent25/src/main/java/org/eclipse/jdt/launching/internal/weaving/ClassfileTransformer2.java
```diff
@@ -0,0 +1,59 @@
...
+		smap.append("*S " + STRATA_ID + "\n"); //$NON-NLS-1$ //$NON-NLS-2$
+		smap.append("*F\n"); //$NON-NLS-1$
+		smap.append("1 ").append(source).append("\n"); //$NON-NLS-1$ //$NON-NLS-2$
+		smap.append("2 ").append(location).append("\n"); //$NON-NLS-1$ //$NON-NLS-2$
+		// JSR-045, StratumSection
+		// "One FileSection and one LineSection (in either order) must follow the StratumSection"
+		smap.append("*L\n"); //$NON-NLS-1$
+		smap.append("*E\n"); //$NON-NLS-1$
+
+		byte[] smapBytes = smap.toString().getBytes(StandardCharsets.UTF_8);
```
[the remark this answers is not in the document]
SougandhS: Handled 👍

SougandhS:
Co-pilot review comments addressed 👍

org.eclipse.jdt.launching/lib/javaagent25.jar
iloveeclipse, line 1: So you add new project but the code is not compiled on build and not automatically updated as jar here.

Please try approach where we don't need to manage a new binary library forever.
Possible options:
- new bundle is a fragment
- optional dependency to new bundle
- whatever else

iloveeclipse (changes requested):

org.eclipse.jdt.launching/lib/javaagent25.jar
SougandhS: I'll check on this 👍

SougandhS:
@iloveeclipse I have added this as an optional dependency as suggested 
Change also requires : https://github.com/eclipse-jdt/eclipse.jdt/pull/169

With these changes I made a local sdk build and was working fine on mac 👍 
Below 25
image


Above 25
image

Jar is now build on eclipse sdk build

PR #998, opened 2026-09-08
Use IFile.getContents() API to retrieve class data in HCR manager

iloveeclipse:
ChangedClassFilesVisitor used ToolFactory.createDefaultClassFileReader(String) API variant to read class bytes for changed class file represented by an IFile. This API uses plain Java.nio code, ignoring Eclipse resources API and possible underlined custom FileSystem/FileStore.

This leads to the problem that for custom file system implementations, which supply file system content on the fly (encoding on write/decoding on read), wrong (not properly decoded) class bytes were read by HCR.

The fix ensures that the IFile change detected by ChangedClassFilesVisitor results in proper class bytes read via IFile.getContents(boolean) API.

Tests contributed by Claude Opus 5.

### DIFF
--- a/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/HcrChangedClassFilesTests.java
+++ b/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/HcrChangedClassFilesTests.java
@@ -1,0 +1,244 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.jdt.debug.tests.core;
+
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.lang.reflect.Method;
+import java.net.URI;
+import java.nio.file.Files;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.eclipse.core.filesystem.URIUtil;
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IFolder;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IProjectDescription;
+import org.eclipse.core.resources.IResource;
+import org.eclipse.core.resources.IResourceChangeEvent;
+import org.eclipse.core.resources.IResourceChangeListener;
+import org.eclipse.core.resources.IWorkspace;
+import org.eclipse.core.resources.IncrementalProjectBuilder;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.Path;
+import org.eclipse.jdt.debug.testplugin.efs.WrapperFileSystem;
+import org.eclipse.jdt.debug.tests.AbstractDebugTest;
+import org.eclipse.jdt.internal.debug.core.hcr.JavaHotCodeReplaceManager;
+
+/**
+ * Tests the detection of changed class files done by the
+ * <code>JavaHotCodeReplaceManager.ChangedClassFilesVisitor</code>, which is the first (and mandatory) step of every hot code replace attempt.
+ * <p>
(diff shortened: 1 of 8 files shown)


org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/HcrChangedClassFilesTests.java
```diff
@@ -0,0 +1,241 @@
...
+			events.incrementAndGet();
+			try {
+				Object visitor = getChangedClassFiles(event);
+				if (visitor == null) {
+					return;
+				}
+				resources.addAll(this.<IResource> invoke(visitor, "getChangedClassFiles")); //$NON-NLS-1$
+				names.addAll(this.<String> invoke(visitor, "getQualifiedNamesList")); //$NON-NLS-1$
+			} catch (Exception e) {
+				error = e;
```
trancexpress: Shouldn't that be a list of errors?

iloveeclipse: Compiler in the IDE says no.

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/HcrChangedClassFilesTests.java
```diff
@@ -0,0 +1,241 @@
...
+	private IFile createProjectWithClassFile(boolean nonLocal) throws Exception {
+		IWorkspace workspace = ResourcesPlugin.getWorkspace();
+		String name = nonLocal ? "HcrNonLocalClassFiles" : "HcrLocalClassFiles"; //$NON-NLS-1$ //$NON-NLS-2$
+		project = workspace.getRoot().getProject(name);
+		if (project.exists()) {
+			project.delete(true, true, null);
+		}
+		IProjectDescription description = workspace.newProjectDescription(name);
+		if (nonLocal) {
+			projectLocation = new File(System.getProperty("java.io.tmpdir"), name + "_" + System.currentTimeMillis()); //$NON-NLS-1$ //$NON-NLS-2$
```
trancexpress: `Files.createTempDirectory()`?

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/HcrChangedClassFilesTests.java
trancexpress: I mean, gather all the errors instead of just keeping the last one?

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/HcrChangedClassFilesTests.java
```diff
@@ -0,0 +1,241 @@
...
+		}
+	}
+
+	/**
+	 * Collects the class files detected by the HCR manager for all POST_BUILD events it receives.
+	 */
+	static class ChangedClassFilesCollector implements IResourceChangeListener {
+
+		final List<IResource> resources = Collections.synchronizedList(new ArrayList<IResource>());
+		final List<String> names = Collections.synchronizedList(new ArrayList<String>());
```
trancexpress: We can skip `<IResource>` and `<String>` in the `synchronizedList` call?

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/HcrChangedClassFilesTests.java
```diff
@@ -0,0 +1,241 @@
...
+				error = e;
+			}
+		}
+
+		/**
+		 * Calls the (protected) <code>JavaHotCodeReplaceManager#getChangedClassFiles(IResourceChangeEvent)</code>, which computes the class files to
+		 * be replaced. Note that the returned visitor type is not accessible, so its (public) methods have to be called reflectively too.
+		 */
+		private static Object getChangedClassFiles(IResourceChangeEvent event) throws Exception {
+			Method method = JavaHotCodeReplaceManager.class.getDeclaredMethod("getChangedClassFiles", IResourceChangeEvent.class); //$NON-NLS-1$
```
trancexpress: IMO it would be better to just test that HCR works also for the special file system case, by having a HCR test with debugging and executing, avoiding the reflection here. But I'm not seeing anything suitable in `HcrTests` to copy...

I guess reflection for protected methods (even internal ones) doesn't hurt that much...

trancexpress (approved):

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/HcrChangedClassFilesTests.java
iloveeclipse: I will just set for the first time

PR #1001, opened 2026-09-09
Allow HCR work if local history is disabled for a source file

iloveeclipse:
In https://github.com/eclipse-platform/eclipse.platform/issues/2588 we added possibility to disable local history for "sensitive" files. However, if such Java file is being modified during debug session, HCR fails because it tries to check whether there was a change by checking the local history entries.

Let assume the file was changed and HCR is needed.

See https://github.com/eclipse-platform/eclipse.platform/issues/2588

Tests contributed by Claude Opus 5.

### DIFF
--- a/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/HcrCompilationUnitDeltaTests.java
+++ b/org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/HcrCompilationUnitDeltaTests.java
@@ -1,0 +1,320 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.jdt.debug.tests.core;
+
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IResource;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.Path;
+import org.eclipse.debug.core.DebugEvent;
+import org.eclipse.debug.core.DebugException;
+import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.IPackageFragment;
+import org.eclipse.jdt.core.IPackageFragmentRoot;
+import org.eclipse.jdt.debug.core.IJavaDebugTarget;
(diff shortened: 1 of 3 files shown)


org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/HcrCompilationUnitDeltaTests.java
```diff
@@ -0,0 +1,323 @@
...
+		@Override
+		public synchronized void obsoleteMethods(IJavaDebugTarget target) {
+			notified = true;
+			notifyAll();
+		}
+
+		synchronized boolean waitNotification() {
+			if (!notified) {
+				try {
+					wait(AbstractDebugTest.DEFAULT_TIMEOUT);
```
trancexpress: Can we use `CountDownLatch` instead of methods on `Object`?

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/HcrCompilationUnitDeltaTests.java
```diff
@@ -0,0 +1,323 @@
...
+		}
+
+		@Override
+		public synchronized void obsoleteMethods(IJavaDebugTarget target) {
+			notified = true;
+			notifyAll();
+		}
+
+		synchronized boolean waitNotification() {
+			if (!notified) {
```
trancexpress: This should be a loop, since `wait()` can return even without the methods here which set `notified=true`.

iloveeclipse: I assume you mean only if we don't use CountDownLatch as proposed above?

org.eclipse.jdt.debug.tests/tests/org/eclipse/jdt/debug/tests/core/HcrCompilationUnitDeltaTests.java
iloveeclipse: Sure

org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/hcr/CompilationUnitDelta.java
```diff
@@ -77,7 +77,12 @@ public CompilationUnitDelta(ICompilationUnit cu, long timestamp)
 
 		// find underlying file
 		IFile file = (IFile) cu.getUnderlyingResource();
-
+		if (file.isContentRestricted()) {
+			// if the file is content restricted, we cannot access its history
+			// but we want HCR to happen anyway, so we assume that the file has changed
+			fHasHistory = true;
```
trancexpress, line 83: Should we set also `fPrevAst`/`fCurrentAst` to `null` here? I see this change will work if they are `null`, but if they are set beforehand (e.g. restriction was added recently) I'm not sure?

iloveeclipse: The "ast" fields are computed in delta constuctor, so "ast" shouldn't be changed. If there was a change on file, then it was made restricted and then yet another change, we should see this in the *other* delta. Or did I misunderstand what you mean?

trancexpress: I thought the object is getting re-used, if new objects are created for each delta then it looks fine.

trancexpress (approved):

